PR #2, opened 2016-10-11
Better attribute binding

OlivierBlanvillain:
@olafurpg If you want to have a look, this I think implements everything we discussed :)

### DIFF
--- a/src/test/scala/BindingTests.scala
+++ b/src/test/scala/BindingTests.scala
@@ -190,21 +190,62 @@ class BindingTests extends FunSuite {
 
   test("onClick = Function0") {
     var clicked = false
-    val button  = <button onclick={ () => clicked = true }>Click me!</button>
+    val button  = <button onclick={ () => clicked = true }>Click Me!</button>
     val div = dom.document.createElement("div")
     mount(div, button)
     assert(!clicked)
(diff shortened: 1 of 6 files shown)


src/test/scala/BindingTests.scala
```diff
@@ -187,4 +187,65 @@ class BindingTests extends FunSuite {
   test("EntityRefMap arrays are equaly sized") {
     assert(EntityRefMap.keys.size == EntityRefMap.values.size)
   }
+
+  test("onClick = Function0") {
```
olafurpg: Is there a way to test the following html

```html
<div class={myClass}>
  <input type="text"/>
</div>
```
and some interaction like this

- myClass is "blue"
- user types in "banana"
- myClass becomes "yellow"
- user input is still "banana"

The most important thing is that the user input is not removed by the class attribute update.

src/test/scala/BindingTests.scala
```diff
@@ -187,4 +187,65 @@ class BindingTests extends FunSuite {
...
+    var clicked = false
+    val button  = <button onclick={ () => clicked = true }>Click Me!</button>
```
olafurpg: Function0, nice 👍 Surprisingly often, you don't care about the type of the event.

PR #5, opened 2016-10-19
Re-invent core value propagation logic

OlivierBlanvillain:
This PR replaces the initial `Monix` based value propagation hack with a new macro free implementation. Hopefully this fixes all memory leak issues.

@Atry If you are interested, https://github.com/OlivierBlanvillain/monadic-html/commit/5d863d8f5ac6974203c5f375c7dce79873844098 is the interesting commit.


Atry:
Looks good

PR #9, opened 2016-10-25
Remove attribute on null.

olafurpg:
To mimic scala-xml behavior, if an attribute value is null, then the
attribute should be removed. The previous behavior was a null pointer
exception.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,13 @@ lib_managed/
 src_managed/
 project/boot/
 project/plugins/project/
+
+# Scala-IDE specific
+.scala_dependencies
+.worksheet
+
+.idea
+
+# ENSIME specific
+.ensime_cache/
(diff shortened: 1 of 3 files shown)


OlivierBlanvillain:
Not a fan of putting semantics into `null`s, but at the same time, I see how this could be useful... LGTM!

olafurpg:
I agree with you on the null, but I think it's best to mimic scala-xml semantics here. It seems you can get the same effect with `Option[T]`, so that would be much much better alternative.

olafurpg:
I was able to achieve the same effect like this:

```scala
    def conditionalAttribute(cond: Boolean): Option[Seq[Node]] =
      if (cond) Some(new Atom("")) else None
<input checked={conditionalAttribute(false)}/>
```

PR #10, opened 2016-10-28
WIP Add demo applications

olafurpg:
My current workflow is

```
cd monadic-examples
browser-sync start --server --files "target/scala-2.11/*.js" # npm install -g browser-sync
sbt "~monadic-examples/fastOptJs"
# hack
```
- [x] Show source file contents in the demos
- [x] Setup github pages


olafurpg:
Okay, I think this is ready now. For the ghpages branch, I think you only need index.html + fullOptJs.

OlivierBlanvillain:
4a920ca moves `index.html` to `/docs`
f9b43ca `git add --force` the output of `examples/fullOpsJS`
WDYT?

OlivierBlanvillain:
That should be enough to get the new `gh-pages` thing working with a single branch... Anyway I don't think we are going to recompile these example so often

olafurpg:
Sure, whatever gets it live on Github pages works :)

olafurpg:
Yay 👏

PR #17, opened 2016-11-02
Doge

olafurpg:


olafurpg:
I couldn't help myself because it's so awesome 
screen shot 2016-11-02 at 22 08 10

OlivierBlanvillain:
wow :open_mouth: 

What about having identical code for this example in [test](https://github.com/OlivierBlanvillain/monadic-html/blob/master/tests/src/test/scala/mhtml/tests.scala#L226-L261) readme and example?

olafurpg:
I pimped up the example with filters since it gives a nice effect to have each "WOW", "MUCH REACTIVE", "SUCH BINDING" message appear one at a time. I'm happy to update the readme example and test to use the same implementation, would that be OK?

OlivierBlanvillain:
:+1

rakajack:
filter seems to be not working as expected, if you decrease the counter, messages "Much reactive" and "Such binding" are not disappearing.

OlivierBlanvillain:
This is working as intended. `filter` is an operation on `Rx` which creates a new `Rx` by "cutting" some the events some from the original reactive variable. It does turns filtered out events into "empty events", you might want to use something like `.map(x => if(f(x)) Some(x) else None` to turn an `Rx[T]` into a `Rx[Option[T]]`.

PR #18, opened 2016-11-06
Fix #14: Implement unsafeRawHTML

### LINKED ISSUE #14: Raw html: dangerouslySetInnerHTML
It would be nice to be able to insert raw html using some syntax like this

```scala
<div>{mhtml.unsafeRawHtml("<h1>Hello!</h1>")}</div>
```

Example use-case: insert html markup from a markdown renderer.

React equivalent: https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml

OlivierBlanvillain:
@olafurpg could you have a look?


olafurpg:
LGTM. I like the name.

PR #19, opened 2016-11-06
Fix #12: Implement .after()

### LINKED ISSUE #12: Support first/per/post update foreach
Or something else equivalent to react component life-cycle.

OlivierBlanvillain:
@olafurpg I could not find a hack to implement this without impacting the core logic, so I hope the added value is worth the complexity. WDYT?


olafurpg:
As we discussed over lunch, I think it's best to put this feature on hold and see if the use-case justifies this complication. I go back to Bergen on Thursday and I will build a bigger application with monadic-html, I'll keep my eyes out for situation where I might want this feature.

OlivierBlanvillain:
Solved with #38 instead

PR #20, opened 2016-11-07
Implement run-time warnings bases on configurable settings	

OlivierBlanvillain:
@olafurpg I still need to add this `-Xelide-below OFF` ScalaCSS thing to toggle configs at compile time, but this should be reviewable as it is.


tests/src/test/scala/mhtml/tests.scala
```diff
@@ -338,4 +336,69 @@ class Tests extends FunSuite {
...
+    assert(w == List("""Implicitly converted class java.lang.Float to it's string representation: "1.1". Call toString explicitly to remove this warning."""))
+  }
+
+  test("EventKey warn") {
+    var w: List[String] = Nil
+    def entity = <div onClick={ () => () }></div>
+    val div = dom.document.createElement("div")
+    mount(div, entity, Config.verbose(s => w = s :: w))
+    assert(div.innerHTML == "<div></div>")
+    assert(w == List("Unknown event onClick. Did you mean onclick instead?"))
```
olafurpg: Awesome! 👏

monadic-html/src/main/scala/mhtml/config.scala
```diff
@@ -0,0 +1,228 @@
...
+    transformAtom           = warnAny2String(warn),
+    transformAttributeValue = warnAny2String(warn),
+    checkElement            = checkElement(warn),
+    checkEventKey           = checkEvent(warn),
+    checkAttributeKey       = checkAttribute(warn)
+  )
+
+  /** Default development configuration; warns about unknown
+    * elements/attributes/entityRefs and implicit Any to String conversions. */
+  val default = verbose(s => println(s"[mhtml] Warning: $s"))
```
olafurpg: Might be good to also provide a `failFast` config that throws errors instead of printing them.

monadic-html/src/main/scala/mhtml/config.scala
```diff
@@ -0,0 +1,228 @@
...
+}
+
+object ConfigHelper {
```
olafurpg: Could we make this a class with members `entityRefKeys`, `events`, `elements`, `attributes`? Then we could do `object ConfigHelper extends ConfigHelper(...)`. This way users could for example easily create a custom config with only a few added entries to `attributes`.

monadic-html/src/main/scala/mhtml/config.scala
```diff
@@ -0,0 +1,228 @@
...
+    val dist = Array.tabulate(s2.length + 1, s1.length + 1) { (j, i) =>
+      if (j == 0) i else if (i == 0) j else 0
+    }
+
+    for(j <- 1 to s2.length; i <- 1 to s1.length)
+      dist(j)(i) =
+        if (s2(j - 1) == s1(i - 1))
+          dist(j - 1)(i - 1)
+        else
+          (dist(j - 1)(i)     + 1) min
```
olafurpg: I don't know what I think of this formatting :p

olafurpg:
Amazing work, I'm really excited to try this out. Only a few minor comments, most importantly on making it easy to create a custom set of attributes/elements/entityrefs.

OlivierBlanvillain:
You can do the following:

```scala
Config.default.copy(checkAttributeKey = att =>
  if (!mySet.contains(att)) System.err.println(s"$att not in $mySet!")
)
```

I guess to expose the levenshtein dist I would need refactor things to use a `Set` instead of `Array` + bin search or people will bump into the sorted requirement on these arrays...

olafurpg:
This new GH review UI is just confusing, it didn't send my review comments because my final comment was send from the "conversations" tab. My only relevant comment is here: https://github.com/OlivierBlanvillain/monadic-html/pull/20#discussion_r86765563

monadic-html/src/main/scala/mhtml/config.scala
OlivierBlanvillain: Throwing stuff in js callbacks won't show in the console (or anywhere) :/

OlivierBlanvillain:
@olafurpg forced pushed a OO version of the `Config`, should be easier to hijack stuff using inheritance. WDYT?

OlivierBlanvillain:
It does! But the diff in file size is very small (3 small arrays and the levenshtein thing), also I was not able to run the tests with `fullOptJS`, apparently the default way does not work in ScalaTest (and I have enough weird ScalaTest stuff to investigate :P)

PR #23, opened 2016-11-16
Bump cats version to 8.1

jac3km4:
I bumped version of cats to 8.1 as well and fixed build errors. I use your custom scala-xml artifacts for both 2.11.8 and 2.12.0, so you will probably want to cross-build it as well.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,9 +1,11 @@
 val scalajsdom = "0.9.1"
 val scalaxml   = "1.0.6"
 val scalatest  = "3.0.0"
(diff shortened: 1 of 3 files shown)


OlivierBlanvillain:
I would like to wait for `scala-xml` to be officially published for 2.12 (and Scala.js) before using 2.12... If you isolate the cats version bump I'm happy to merge it!

OlivierBlanvillain:
LGTM :+1:, many thanks :-)

PR #25, opened 2016-11-22
Explain precise data-binding mechanism in README

OlivierBlanvillain:
@olafurpg WDYT?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ val view =
   </div>
 ```
 
-When mounting this view, [the implementation](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-html/src/main/scala/mhtml/mount.scala) with attach callbacks to each `Rx` such that changing `a`, `b` or `c` results in very precise DOM updates:
+When mounting this view, [the implementation](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-html/src/main/scala/mhtml/mount.scala) will attach callbacks to each `Rx` such that changing `a`, `b` or `c` results in precise DOM updates:
 
 - Changing `a` will update the `div` attribute (reusing the same `div` node)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,7 +72,32 @@ class Var[A](initialValue: A) extends Rx[A] {
...
+val b = Var("foo")
+val c = Var("bar")
+
+val view =
+  <div id={a}>
+    Variable 1: {b}; variable 2: {c}.
+  </div>
+```
+
+When mounting this view, [the implementation](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-html/src/main/scala/mhtml/mount.scala) with attach callbacks to each `Rx` such that changing `a`, `b` or `c` results in very precise DOM updates:
```
olafurpg: s/with/will/

s/very precise/the most possible precise/

README.md
```diff
@@ -72,7 +72,32 @@ class Var[A](initialValue: A) extends Rx[A] {
...
+
+- Changing `a` will update the `div` attribute (reusing the same `div` node)
+- Changing `b` will delete the text node between `Variable 1: ` and `; variable 2: `, and insert a new replacement between these two nodes.
+- Changing `c` will do the same, except between `; variable 2: ` and `.`.
+
+These updates correspond to what React would compute using virtual-DOM diffing.
+
+When working with large immutable data structures, this approach is less performant that virtual-DOM diffing. Indeed, creating a large view out of a `Rx[List[_]]` implies that any changes to the `List` triggers a re-rendering of the entirety of the view. We plan to address this point in [#13](https://github.com/OlivierBlanvillain/monadic-html/issues/13) by combining the current approach with targeted virtual-DOM.
```
olafurpg: s/that/than/

README.md
olafurpg: Or just s/very//

I don't like "very"

Issue #16, opened 2016-11-02
Index out of bounds error


olafurpg:
```scala
    val count: Var[Int] = Var[Int](0)

    val dogs: Rx[Seq[xml.Node]] =
      count.map(i =>
        Seq.fill(i)(
          <img style="width: 100px;" src="http://doge2048.com/meta/doge-600.png"></img>))
    count.filter(_ > 0).foreach(println)

    <div style="background-color: darkorange;">
      <button onclick={ () => count.update(_ + 1) }>Click Me!</button>
      {count.filter(_ > 0).map(<h2>WOW!!!</h2>)}
      {count.filter(_ > 2).map(<h2>MUCH REACTIVE!!!</h2>)}
      {count.filter(_ > 5).map(<h2>SUCH BINDING!!!</h2>)}
      {dogs}
    </div>
```

Produces the following error
```
hd	@	LinearSeqOptimized.scala:65
c.la	@	List.scala:84
c.j	@	NodeSeq.scala:55
(anonymous function)	@	rx.scala:14
Bo.j	@	AnonFunctions.scala:15
Qb	@	rx.scala:77
(anonymous function)	@	rx.scala:18
Bo.j	@	AnonFunctions.scala:15
Qb	@	rx.scala:77
(anonymous function)	@	rx.scala:83
Bo.j	@	AnonFunctions.scala:15
Lb	@	rx.scala:64
yk	@	rx.scala:83
Pk.hb	@	Dogs.scala:16
(anonymous function)	@	mount.scala:76
Bo.j	@	AnonFunctions.scala:15
(anonymous function)	@	mount.scala:87
```

The culprit line is here: https://github.com/OlivierBlanvillain/monadic-html/blob/958184f2e81f042ece08981fe6043ed3c3c35012/monadic-rx/shared/src/main/scala/mhtml/rx.scala#L14

I'm surprised it even typechecks because I don't pass in a function into the map. When I pass a function (`.map(_ => <div>...</div>)`) it works.

OlivierBlanvillain:
Bumped into this one already, here is the [root cause](https://github.com/scala/scala-xml/blob/892c2a7/src/main/scala/scala/xml/NodeSeq.scala#L55) :/

[...]

PR #26, opened 2016-12-02
Fix scala-xml

OlivierBlanvillain:
@olafurpg Could you have a look? Can you also test it on your project to see if things break, and how it impacts the size of `fullOptJS`?


olafurpg:
Are those 100 lines of code enough to replace scala-xml?



I won't miss `NodeSeq` or `Node <:!< Seq[Node]` (total wtf, causes bugs like #16).

I compiled and ran my project on a local publish of this PR and can confirm that 1. it seems to work and 2. fullOptJs is 30 KB smaller. Good job!

I am surprised the size decrease isn't more dramatic. I guess Scala.js code eliminnation is already so good! (kudos @sjrd) The great simplification of scala-xml-light over scala-xml is very welcome and reason enough to merge this PR. LGTM 👍

OlivierBlanvillain:
> Are those 100 lines of code enough to replace scala-xml?

Yes. `scalac` apparently parses xml literals into `scala.xml._` stuff, anything with same names and properly shaped constructors can be used instead. This should be close to the minimum subset we need for our use-case :). It's also probably going to make implementing #13 a lot cleaner.

PR #27, opened 2016-12-03
Restrict which types are allowed to be embedded in xml.

olafurpg:
This commit changes the "Implicitly converted X to its String representation"
runtime warnings into a compiler error. The following error appears when a user inserts an invalid value in xml element position
```
Cannot embed value of type A in xml element, implicit XmlElementEmbeddable not found.
The following types are supported:
- String, Int, Long Double, Float, Char, Boolean
- xml.Node, Seq[xml.Node], List[xml.Node] (or any iterable container of xml.Node)
- mhtml.Var[T], mhtml.Rx[T] where T can be embedded in xml element position.
- Option[T] where T can be embedded in xml element position.
[error]     def entity = <div>{new A}</div>
```

This change eliminates the need for implicit conversions to `xml.Node` in the
mhtml package object and the need for `import mhtml._` in most places.

The scala-xml-light project is merged into the monadic-html project, in order to
include mhtml.Rx as part of the xml api.


olafurpg:
After some thinking, we might want to replace `Show[T]` with `XmlSerializable[T]` (or some similar name). The trait would have no method. The existence of an implicit `XmlSerializable[T]` is a proof that the value is allowed to be inside an xml literal. What operations you can do on such an instance is up to you to define.

The big benefit to this type-class is that it helps you from accidentally inserting unknown values into xml literals. I don't think the main value is customizing the string representation of my domain objects inside xml literals, I am OK with using the current `toString` approach.

OlivierBlanvillain:
It's very interesting that this works :). Now the question is, is this better that the run-time warning? You get better error locality, but it's "harder" to solve: the error suggests to declare a implicit in scope instead of calling `.toString`.

I think we could get the best of both world with something like

```scala
@implicitNotFound("This XML literal using a value of type ${A} in a template position, call `.toString` to explicitly.")
sealed trait XmlSerializable[A] // It's sealed

object XmlSerializable {
  implicit val string: XmlSerializable[String] = new XmlSerializable[String] {}
  implicit val rx:     XmlSerializable[String] = new XmlSerializable[Rx[String]] {}
}
```

olafurpg:
Good point. I like the idea of making the trait sealed and customizing the `implicitNotFound` error with clear instructions on what values can be embedded in xml literals. Note that not only strings are allowed, but also `Function0[Unit]`, `Function1[A, Unit]`, `Node` and `Iterable[Node]`.

I'll try to patch up this PR asap, but I may not have time for it until after my scalax talk next week Friday. I think this change in combination with scala-xml-light warrant a bump to v0.2 :)

olafurpg:
Actually, if we're gonna make the trait sealed, we might as well skip the type class and hard-code the allowed types into `NodeBuffer.&+` using method overloading. I'm not sure if results in cleaner compiler errors or faster compilation, but it's worth a thought. WDYT?

OlivierBlanvillain:
Agreed :+1:

olafurpg:
What do you think about these error messages:

```
mhtml/tests.scala:357: overloaded method value &+ with alternatives:
[error]   [A](f: A => Unit)scala.xml.NodeBuffer <and>
[error]   (f: () => Unit)scala.xml.NodeBuffer <and>
[error]   (e: mhtml.Rx[String])scala.xml.NodeBuffer <and>
[error]   (e: String)scala.xml.NodeBuffer <and>
[error]   (it: Iterable[scala.xml.Node])scala.xml.NodeBuffer <and>
[error]   (node: scala.xml.Node)scala.xml.NodeBuffer
[error]  cannot be applied to (Double)
[error]     def entity = <div>{1.1}</div>

**Comment by olafurpg** [comment]
I updated this PR and commit message following your feedback.

File: monadic-html/src/main/scala/mhtml/package.scala [hunk 1]
```diff
@@ -1,12 +1,7 @@
-import scala.language.implicitConversions
 
 import scala.xml.Node
 
 package object mhtml {
-  implicit def RxToOSN[A](a: Rx[A]): Node                     = new xml.Atom(a)
```
OlivierBlanvillain: Sweet! Removing these implicit conversions is IMO enough motivation to move forward with this :smile:

tests/src/test/scala/mhtml/tests.scala
```diff
@@ -381,11 +372,10 @@ class Tests extends FunSuite {
 
   test("README warnings") {
     var w: List[String] = Nil
-    def entity = <captain yolo="true" onClick={ () => println("Oh yeah!") }>{None}</captain>
+    def entity = <captain yolo="true" onClick={ () => println("Oh yeah!") }>{None.toString}</captain>
     val div = dom.document.createElement("div")
     mount(div, entity, new DevSettings { override def warn(s: String) = w = s :: w })
     assert(w == List(
-      """Implicitly converted class scala.None$ to it's string representation: "None". Call toString explicitly to remove this warning.""",
```
OlivierBlanvillain: We should probably also remove the infrastructure to generate this warning.

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -76,27 +76,51 @@ final case class PrefixedAttribute(
...
 
-final case class UnprefixedAttribute(
+final case class UnprefixedAttribute[T] private (
   key: String,
   value: Node,
   next: MetaData
 ) extends MetaData {
-  def this(key: String, value: String, next: MetaData) =
-    this(key, Text(value), next)
+  def this(key: String, e: T, next: MetaData)(implicit ev: XmlSerializable[T]) =
```
OlivierBlanvillain: Did you try making this the primary constructor?

olafurpg:
I added another commit that uses the XmlSerializable trait to re-use Node creation logic. It's not necessary, but I think it makes the code quite clean.

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -76,27 +76,55 @@ final case class PrefixedAttribute(
...
 
-// Internal structure used by scalac to create literals -----------------------
+@scala.annotation.implicitNotFound(msg =
+  """Cannot embed value of type '${T}' into an xml literal. Allowed types are:
+Base types:
+  - String (tip: you can use ${T}.toString)
+  - Int
+  - xml.Node
+  - List[xml.Node], Seq[xml.Node] and other Iterable containers.
+Event handlers:
```
OlivierBlanvillain: For example with two type classes, `XmlElementEmbeddable` and `XmlAttributeEmbeddable`

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -76,27 +76,49 @@ final case class PrefixedAttribute(
...
+  - () => Unit
+  - T => Unit, where T can be any type.
+Monadic-html:
+  - mhtml.Rx[T] or mhtml.Var[T], where T is a type that can be embedded into an xml literal. For example, a String or Seq[xml.Node].
+""")
+sealed trait XmlSerializable[T] { def toNode(e: T): Node }
```
OlivierBlanvillain: I think it could be nice to keep `scala-xml-light` and `mhtml` separated if possible, it would at least look less hacky from outside to be able to say "hey looks this is a light, scalac compatible alternative to `scala-xml`" and not something deeply embedded into the lib...

OlivierBlanvillain: By that I mean something like a non sealed `XmlSerializable` with instances inside `package object mhtml`

monadic-html/src/main/scala/scala/xml/xml.scala
olafurpg: Nope, didn't think of that 😄

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -76,27 +76,55 @@ final case class PrefixedAttribute(
...
+      next
+    )
 }
 
-// Internal structure used by scalac to create literals -----------------------
+@scala.annotation.implicitNotFound(msg =
+  """Cannot embed value of type '${T}' into an xml literal. Allowed types are:
+Base types:
+  - String (tip: you can use ${T}.toString)
+  - Int
```
olafurpg: I agree, but I'm still torn on this one because it's so convenient to not have to do `.toString` on numbers. How about we include all literal types? Int, Double, Long, Float.

monadic-html/src/main/scala/scala/xml/xml.scala
olafurpg: I like where you're going. This is totally doable.

monadic-html/src/main/scala/scala/xml/xml.scala
olafurpg: I agree we should avoid coupling scala-xml-light with mhtml. I moved it in order to keep the trait sealed, but I suppose we're safe to unseal it and move the implicit instance into mhtml package object.j

tests/src/test/scala/mhtml/tests.scala
olafurpg: Agreed!

monadic-html/src/main/scala/scala/xml/xml.scala
OlivierBlanvillain: Alright :+1:

olafurpg:
OK, 3rd attempt down and I'm quite happy with this now. Let me know what you think!

examples/src/main/scala/mhtml/examples/HelloWorld.scala
```diff
@@ -1,4 +1,5 @@
 package mhtml.examples
+import mhtml._
```
OlivierBlanvillain: Uh this is bad, it means that first experience with the lib is going to be `error: could not find implicit value for parameter ev: XmlAttributeEmbeddable[String]`, which cannot really be made nice because `@implicitNotFound` needs to be defined locally. Maybe we should forget about the idea of splitting `xml-light` appart... WDYT? Sorry for the back and forth...

tests/src/test/scala/mhtml/tests.scala
```diff
@@ -352,15 +352,6 @@ class Tests extends FunSuite {
     assert(w == List("Unknown element yolo. Did you mean col instead?"))
   }
 
-  test("Atom warn") {
```
OlivierBlanvillain: Do you think you could make a few negative test cases with [assertTypeError](http://doc.scalatest.org/3.0.0/index.html#org.scalatest.Assertions@assertTypeError(code:String)(implicitpos:org.scalactic.source.Position):org.scalatest.Assertion)?

olafurpg: Great idea, I didn't know scalatest had that.

examples/src/main/scala/mhtml/examples/HelloWorld.scala
olafurpg: I know, it's a shame. How about we customize the `implicitNotFound` message in the `scala-xml-light` project to the previous message:

```
Cannot embed value of type ${T} in xml element. Here's what you can try:
1. `import mhtml._`
2. Use `${T}.toString`
3. Find which part of ${T} is disallowed. Only the following types are allowed
  - String, Int, Long Double, Float
  - xml.Node, Iterable[xml.Node]
  - mhtml.Var[T], mhtml.Rx[T] where T can be embedded in html
```

This means only the error message (which is easy to move) is tied to mhtml while scala-xml.light remains a separate module.

olafurpg:
BTW. I found some unexpected behavior by putting iterables into attributes.

```
// this
<form disabled={Option(Text("string"))}></form>
// causes runtime error
[info]   scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: Group(List(Text(string))) is not an instance of scala.xml.Atom
```

I need more time to investigate this.

OlivierBlanvillain:
Does it even makes sense to have iterables in attribute position? (I mean, should this even compile?)

olafurpg:
Nope, it doesn't make sense. I need to fine-tune the Attribute implicits.

examples/src/main/scala/mhtml/examples/HelloWorld.scala
OlivierBlanvillain: I don't know, I think it's weird to have to import `mhtml._` to use `scala.xml._`, putting them in the companion object removes the need for wild card import entirely, so I would say "fuck it", we are too deep into a hack to care about modularity :P

OlivierBlanvillain: I would go for something like this:

```scala
Cannot embed value of type ${T} in XML element. The following types are supported:
  - String, Int, Long, Double, Float // Char, Boolean (?)
  - xml.Node, Iterable[xml.Node] // Should we use `Seq` instead?, should we allow Seq[String]?
  - mhtml.Rx[T] where T is itself supported // Var implied with sub-typing
```

olafurpg: My idea was to tailor the error message to a less experienced audience, for example frontend developers who are taking their first steps in Scala. That's why I thought the "troubleshooting" part might be smart and to explicitly mention that both Var and Rx are OK.

Veteran Scala developers should have an easier time figuring out what's happening without help from the error message.

I agree that Char and Boolean should be included, to support all literal types.
I also agree that we should say Seq[xml.Node] instead of Iterable[xml.Node]

olafurpg: Wait, mhtml._ import is not necessary anymore since the implicits are defined in the companion objects. Cool! I'll go without the 3 step troubleshooting guide but keep the allowed types quite explicit (e.g., mention Rx AND Var).

olafurpg:
Okie dokie, attempt nr. 4. This PR is improving every time!

Although it's not strictly related to this PR, I took the liberty to define Option[T] semantics in element and attribute position, fixing #29. I included it here since I was doing the well-typed tests and I didn't like the previous behavior.

olafurpg:
I fixed the "unused import" compilation errors in the examples (hurray, no `import mhtml._` required anymore!) and added a section in the readme.

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,154 @@
...
+  override val value: Node = ev.toNode(e)
+}
+
+/** Evidence that T can be embedded in xml attribute position. */
+@scala.annotation.implicitNotFound(msg =
+    """Cannot embed value of type ${T} in xml attribute, implicit XmlAttributeEmbeddable not found.
+The following types are supported:
+- String, Boolean
+- xml.Text
+- Option[String]
```
OlivierBlanvillain: Why not `Option[T] where T can be embedded in xml attribute position.` here as well?

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,154 @@
...
+  implicit val intElementEmbeddable: XmlElementEmbeddable[Int]           = atom[Int]
+  implicit val floatElementEmbeddable: XmlElementEmbeddable[Float]       = atom[Float]
+  implicit val doubleElementEmbeddable: XmlElementEmbeddable[Double]     = atom[Double]
+  implicit val longElementEmbeddable: XmlElementEmbeddable[Long]         = atom[Long]
+  implicit val charElementEmbeddable: XmlElementEmbeddable[Char]         = instance[Char](x => Text(x.toString))
+  implicit val stringElementEmbeddable: XmlElementEmbeddable[String]     = instance[String](Text.apply)
+  implicit def nodeElementEmbeddable[T <: Node]: XmlElementEmbeddable[T] = instance[T](identity)
+  implicit def iterableElementEmbeddable[C[_], T <: Node](implicit
+      ev: XmlElementEmbeddable[T],
+      conv: C[T] => Iterable[T]
```
OlivierBlanvillain: Do we actually are about any `Iterable[T]` which is not `<: Seq[T]`? If not I would just use something like the following:

```scala
implicit def seqElementEmbeddable[T <: Node]: XmlElementEmbeddable[Seq[T]] =
  instance[Seq[T]](x => Group(x))
```

tests/src/test/scala/mhtml/tests.scala
```diff
@@ -410,4 +406,45 @@ class Tests extends FunSuite {
...
+    assertTypeError("<form disabled={Option.empty[Int]}></form>")
+    assertTypeError("<form disabled={List.empty[String]}></form>")
+    assertTypeError("<form disabled={List.empty[Text]}></form>")
+    assertTypeError("<form disabled={Text(\"\"): xml.Node}></form>")
+    assertTypeError("<form disabled={Rx(<div></div>)}></form>")
+  }
+
+  test("ill-typed Element") {
+    assertTypeError("<div>{true}</div>")
+    assertTypeError("<div>{List.empty[String]}</div>")
```
OlivierBlanvillain: hmm shouldn't these two compile?

monadic-html/src/main/scala/mhtml/mount.scala
```diff
@@ -94,15 +98,18 @@ object mount {
 
     def setMetadata(m: MetaData, v: Any, config: MountSettings): Unit = {
       val htmlNode = node.asInstanceOf[dom.html.Html]
-      def set(key: String): Unit =
-        if (v == null) htmlNode.removeAttribute(key)
-        else {
+      def set(key: String): Unit = v match {
+        case null | None | false => htmlNode.removeAttribute(key)
```
OlivierBlanvillain: TBH I'm not really sold on this `true`/`false` sementic, what if the user wants to actually print false? Isn't a bit weird to have `1 → "1"`, `1.1 → "1.1"`, `'c' → "c"` but `false → Ø`?

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,154 @@
...
+  ): XmlElementEmbeddable[C[T]] =
+    instance[C[T]](x => Group(conv(x).map(ev.toNode).toSeq))
+  implicit def optionElementEmbeddable[T](implicit ev: XmlElementEmbeddable[T]): XmlElementEmbeddable[Option[T]] =
+    instance[Option[T]] {
```
OlivierBlanvillain: I wonder if this should wrap the `Option` in a `Atom` and let mounting to the magic bits

olafurpg:
No worries, this is a big change and the PR has improved every round of review : )

I'm busy preparing for scala exchange this week. I'll get back to this next week!

tests/src/test/scala/mhtml/tests.scala
```diff
@@ -410,4 +406,41 @@ class Tests extends FunSuite {
...
+    assert(render(<div>{Option.empty[Int]}</div>) == "<div></div>")
+    assert(render(<div>{Rx(Option(1))}</div>) == "<div>1</div>")
+    assert(render(<div>{Option(1)}</div>) == "<div>1</div>")
+    assert(render(<div>{1}</div>) == "<div>1</div>")
+    assert(render(<div>{2L}</div>) == "<div>2</div>")
+    assert(render(<div>{1.2d}</div>) == "<div>1.2</div>")
+    assert(render(<div>{1.0f}</div>) == "<div>1</div>")
+    assert(render(<div>{"string"}</div>) == "<div>string</div>")
+    assert(render(<div>{List(scala.xml.Comment("a"))}</div>) == "<div><!--a--></div>")
+    assert(render(<div>{List(scala.xml.Comment("a"))}</div>) == "<div><!--a--></div>")
```
olafurpg: Good catch. I've refactored these tests into a separate test suite so that each assertion becomes a test with the name of the expected output. This should help catch duplicate tests in the future.

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/** Evidence that T can be embedded in xml attribute position. */
+@scala.annotation.implicitNotFound(msg =
+    """Cannot embed value of type ${T} in xml attribute, implicit XmlAttributeEmbeddable not found.
+The following types are supported:
+- String, Boolean
+- xml.Text
+- Option[String]
+- () => Unit, T => Unit event handler. Note. The return type needs to be Unit!
```
olafurpg: Done.

OlivierBlanvillain:
For this `XmlElementEmbeddable[Seq[T]]` thing, you could try with `C[_] <: Seq[_]`:

```scala
scala> trait E[T]
defined trait E

scala> implicit val aaa: E[String] = new E[String] {}
aaa: E[String] = $anon$1@546621c4

scala> implicit def i[C[_] <: Seq[_], T](implicit ev: E[T]): E[C[T]] = null
i: [C[_] <: Seq[_], T](implicit ev: E[T])E[C[T]]

scala> implicitly[E[Seq[String]]]
res0: E[Seq[String]] = null

scala> implicitly[E[List[String]]]
res1: E[List[String]] = null
```

OlivierBlanvillain:
> Does `mount` support, for example, `Option[Rx[() => Unit]]`?

With a `case Some(v)` in [this pattern matching](https://github.com/OlivierBlanvillain/monadic-html/blob/efde1d44c0a00ba748742f2d06c488e79a1a7986/monadic-html/src/main/scala/mhtml/mount.scala#L70-L85) I think this might be possible.

OlivierBlanvillain:
About the attributes/elements thing, do you think we could sum it all up like this in the doc:?

#### Both attributes and elements:

- `String`
- `mhtml.Var[T], mhtml.Rx[T] where T is itself embeddable`
- `Option[T] where T can itself be embedded (None → remove from the DOM)`

#### Attributes

- `Boolean (false → remove from the DOM)`
- `() => Unit, T => Unit event handler`
- `xml.Text` // Do we have a use case of this one?

#### Elements

- `Int, Long, Double, Float, Char, Boolean (silently converted to string)`
- `xml.Node`
- `Seq[xml.Node]`

olafurpg:
Here we go!

- I had tried `C[_] <: Seq[_]` but it always required a cast, for some reason. I've updated the PR removing the view bound in favor of a context bound for `C[_] <: Seq[_]` but using the cast (tests pass, would be interested to know what case causes classcastexception).
- Handle Some(x) in mount.
- Updated readme with your list, removed xml examples.

monadic-html/src/main/scala/mhtml/mount.scala
```diff
@@ -94,15 +100,18 @@ object mount {
...
+        case _ =>
           config.inspectAttributeKey(key)
-          // This should be a `config.transformAttributeValue`, but scala-xml does not compile non String attribues...
-          val value = v.toString
+          val value = v match {
+            case Some(x) => x.toString
```
OlivierBlanvillain: This case should never append now that you've added the `case Some` in the other patmat

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,149 @@
...
+  implicit def function1AttributeEmbeddable[T]: XmlAttributeEmbeddable[T => Unit] = atom[T => Unit]
+  implicit def optionStringAttributeEmbeddable[T:XmlAttributeEmbeddable]: XmlAttributeEmbeddable[Option[T]]        = atom[Option[T]]
+  implicit def rxAttributeEmbeddable[C[_] <: mhtml.Rx[_], T: XmlAttributeEmbeddable]: XmlAttributeEmbeddable[C[T]] = atom[C[T]]
+}
+
+/** Evidence that T can be embedded in xml element position. */
+@scala.annotation.implicitNotFound(msg =
+    """Cannot embed value of type ${T} in xml element, implicit XmlElementEmbeddable[${T}] not found.
+The following types are supported:
+- String, Int, Long, Double, Float, Char, Boolean
```
OlivierBlanvillain: I think Boolean should be removed here

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,149 @@
...
+- xml.Node, Seq[xml.Node]
+- mhtml.Var[T], mhtml.Rx[T] where T can be embedded in xml element position.
+- Option[T] where T can be embedded in xml element position.
```
OlivierBlanvillain: Might want to add `(None → remove from the DOM)` to make it extra clear...

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -0,0 +1,149 @@
...
+)(implicit ev: XmlAttributeEmbeddable[T]) extends MetaData {
+  override val value: Node = ev.toNode(e)
+}
+
+/** Evidence that T can be embedded in xml attribute position. */
+@scala.annotation.implicitNotFound(msg =
+    """Cannot embed value of type ${T} in xml attribute, implicit XmlAttributeEmbeddable[${T}] not found.
+The following types are supported:
+- String, Boolean
+- xml.Text
```
OlivierBlanvillain: To be removed

OlivierBlanvillain:
I'm really confused by this github review stuff..

Anyway final nitpicks and lets merge :+1:. Congrats on this awesome hack! :)

monadic-html/src/main/scala/scala/xml/xml.scala
olafurpg: Done.

monadic-html/src/main/scala/scala/xml/xml.scala
olafurpg: Good catch! Done.

monadic-html/src/main/scala/scala/xml/xml.scala
olafurpg: Agreed.

monadic-html/src/main/scala/mhtml/mount.scala
olafurpg: It is needed since this is case handles element position and the other case handles attribute position.

OlivierBlanvillain:
I checked-out your branch to try out what I did on the first commit:

> In [your latest comment](https://github.com/OlivierBlanvillain/monadic-html/pull/27/files/ea07d84408ba6e2ef174ff3a4e69c8c5b71dbe7f#r92760654) I think you are confusing [this block in the body of `mountNode`](https://github.com/olafurpg/monadic-html/blob/e9f815fb1e1da30945298392ef2ffc9a76ff4ef3/monadic-html/src/main/scala/mhtml/mount.scala#L55-L59) with the one I'm pointing at [in the body of `setMetadata`](https://github.com/olafurpg/monadic-html/blob/e9f815fb1e1da30945298392ef2ffc9a76ff4ef3/monadic-html/src/main/scala/mhtml/mount.scala#L107-L111), this is the change I had in mind...

And then changed to stuff to fix `<div>{Some(<p></p>)}</div>` which was first not compiling, and then not rendering as expected :)

olafurpg:
LGTM 👍 Thanks for polishing. I agree, the gh review ui is confusing for large PR like this

OlivierBlanvillain:
:fireworks:

olafurpg:
Whoop whoop 🍾

PR #28, opened 2016-12-04
Restrict which types are allowed to be embedded in xml.

olafurpg:
This commit changes the "Implicitly converted X to it's string, representation"
runtime warnings into a compiler error. A clean compiler error message should
appear when invalid types are inserted into xml literals. Example:

```
test/scala/mhtml/XXX.scala:10: Cannot embed value of type 'A' into an xml literal. Allowed types are:
[error] Base types:
[error]   - String (tip: you can use A.toString)
[error]   - Int
[error]   - xml.Node
[error]   - List[xml.Node], Seq[xml.Node] and other Iterable containers.
[error] Event handlers:
[error]   - () => Unit
[error]   - T => Unit, where T can be any type.
[error] Monadic-html:
[error]   - mhtml.Rx[T] or mhtml.Var[T], where T is a type that can be embedded into an xml literal. For example, a String or Seq[xml.Node].
[error]     def entity = <div>{new A}</div>
```

This change eliminates the need for implicit conversions to `xml.Node` in the
mhtml package object.

The scala-xml-light project is merged into the monadic-html repo, in order to
include mhtml.Rx as part of the xml api.


olafurpg:
Any suggestions on the `implicitNotFound` message? It's a bit verbose for my taste, but tries to be friendly and helpful.

olafurpg:
Whoops, closing this in favor of existing PR #27

PR #32, opened 2016-12-19
Remove Rx#filter and refactor implementation accordingly

OlivierBlanvillain:
Also adds `Rx#foreachNext` and `Rx#get`, not sure it's wise to have them in public api...

@olafurpg what do you think?

### DIFF
--- a/examples/src/main/scala/mhtml/examples/Chosen.scala
+++ b/examples/src/main/scala/mhtml/examples/Chosen.scala
@@ -105,10 +105,15 @@ object Chosen {
       div -> candidates
     }
     val rxCandidates: Rx[Seq[T]] = rxCandidatesWithApp.map(_._2)
-    val highlightedCandidate: Rx[T] =
+    val highlightedCandidate: Rx[T] = {
+      val filtered = Var[T](rxCandidates.get.head)
+
       (for { index <- rxIndex; candidates <- rxCandidates } yield {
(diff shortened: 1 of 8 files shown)


examples/src/main/scala/mhtml/examples/Chosen.scala
```diff
@@ -105,10 +105,15 @@ object Chosen {
...
     }
     val rxCandidates: Rx[Seq[T]] = rxCandidatesWithApp.map(_._2)
-    val highlightedCandidate: Rx[T] =
+    val highlightedCandidate: Rx[T] = {
+      val filtered = Var[T](rxCandidates.get.head)
+
       (for { index <- rxIndex; candidates <- rxCandidates } yield {
         candidates.zipWithIndex.find(_._2 == index).map(_._1)
-      }).collect { case Some(x) => x }
+      }).foreach(_.foreach(filtered.:=))
```
olafurpg: This looks a bit weird but I'm sure I would have written it differently if `collect` hadn't existed.

OlivierBlanvillain:
I think `get` should be scary, I would consider it as bad a Option#get, i.e. if you use it for anything but tests you are doing it wrong...

olafurpg:
but will `.get` ever throw an exception? If not, then I think we should use a different name since `Option.get` does throw an exception. Maybe `underlyingValue`?

OlivierBlanvillain:
Alright I'm sold on `.value` :smile:

PR #36, opened 2017-01-23
add namespace to node and use createElementNS if present

adrobisch:
some nodes like svg will only work properly with createElementNS
so the namespace is needed during node creation. this is a small patch to fix that

### DIFF
--- a/monadic-html/src/main/scala/scala/xml/xml.scala
+++ b/monadic-html/src/main/scala/scala/xml/xml.scala
@@ -12,8 +12,7 @@ import language.higherKinds
 sealed trait Node {
   def scope: Option[Scope] = None
   def prefix: Option[String] = None
-  def namespace: Option[String] = if (prefix.isDefined) prefix.flatMap(namespaceURI) else scope.flatMap(_.namespaceURI(null))
-  def namespaceURI(prefix: String): Option[String] = scope.flatMap(_.namespaceURI(prefix))
+  def namespace: Option[String] = scope.flatMap(_.namespaceURI(prefix.orNull))
 }
 
 /** A hack to group XML nodes in one node. */
@@ -99,8 +98,6 @@ sealed trait MetaData extends Iterable[MetaData] {
 
   def iterator: Iterator[MetaData] =
     Iterator.single(this) ++ next.iterator
-
-  def namespace: Option[String] = value.namespace
 }
 
 case object Null extends MetaData {
@@ -117,8 +114,6 @@ final case class PrefixedAttribute[T](
   next: MetaData
 )(implicit ev: XmlAttributeEmbeddable[T]) extends MetaData {
   override val value: Node = ev.toNode(e)
-
-  override def namespace: Option[String] = value.namespaceURI(pre)
 }
 
 final case class UnprefixedAttribute[T](
--- a/monadic-html/src/main/scala/scala/xml/xml.scala
+++ b/monadic-html/src/main/scala/scala/xml/xml.scala
@@ -18,6 +18,7 @@ final case class Group(nodes: Seq[Node]) extends Node
 final case class Elem(
   label: String,
   attributes1: MetaData,
+  namespace: Option[String],
   child: Node*
 ) extends Node {
   // m, former minimizeEmpty, and p former prefix are now thrown away.
@@ -32,6 +33,11 @@ final case class Elem(
         case _ => acc
       }
       merge(a, s)
+    }, s match {
+      case TopScope => None
+      case NamespaceBinding(null, url, next) => Some(url)
+      case NamespaceBinding(key, url, next) => Some(url)
+      case _ => None
     }, c: _*)
 }
 
--- a/monadic-html/src/main/scala/scala/xml/xml.scala
+++ b/monadic-html/src/main/scala/scala/xml/xml.scala
@@ -9,21 +9,27 @@ import language.higherKinds
 // directly in scalac paser). `NoteSeq` disappeared, `Node <:!< Seq[Node]`...
 
 /** Trait representing XML tree. */
-sealed trait Node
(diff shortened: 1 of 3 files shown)


OlivierBlanvillain:
This is more complicated than I expected, I guess I should've read the spec for namespaces :smiley:. We definitely need to revert parts of 91cff868849406ab7c6ee41b414dd42e1fb6931e.

It looks like in addition to [createElementNS](https://github.com/lihaoyi/scalatags/blob/3dea48c42c5581329e363d8c3f587c2c50d92f85/scalatags/js/src/main/scala/scalatags/JsDom.scala#L180) ScalaTags also uses [setAttributeNS](https://github.com/lihaoyi/scalatags/blob/3dea48c42c5581329e363d8c3f587c2c50d92f85/scalatags/js/src/main/scala/scalatags/JsDom.scala#L150), I guess we also need to use that instead of creating `foo:bar` attributes like we do now?

adrobisch:
@olafurpg I added tests for the creation with and w/o namespace. Is that sufficient?
@OlivierBlanvillain Yes, for prefixed attributes the namespace should be looked up in the scopes and if there is one `setAttributeNS` should be used. Currently there is no way to get the namespace from a scope (or its parent), which would be needed so that the attribute can lookup the prefix namespace in the owning node. Do you want to continue on this (since you mentioned a revert)? I could also look into that with this PR.

OlivierBlanvillain:
@adrobisch My understanding is that generating the following SVG:

```xml
<svg xmlns="http://www.w3.org/2000/svg"
     xmlns:xlink="http://www.w3.org/1999/xlink">
</svg>
```

Requires something like:

```scala
val elemNode = document.createElementNS("http://www.w3.org/2000/svg", "svg")
elemNode.setAttributeNS("http://www.w3.org/2000/xmlns", "xmlns:xlink", "http://www.w3.org/1999/xlink")
```

For the SVG code snippet above, I think the scalac XML parser ends up calling into scala.xml with something like the following:

```scala
Elem(
  previx   = "",
  label    = "svg",
  metadata = Null,
  scope    = NamespaceBinding(null, "http://www.w3.org/2000/svg", 
               NamespaceBinding("xlink", "http://www.w3.org/1999/xlink", null)
             ),
  minimize = false,
  child    = Nil
)
```

The pattern matching you added [here](https://github.com/OlivierBlanvillain/monadic-html/pull/36/files#diff-22ee1db36eb09cacb3700f66645f7043R38) should not throws away the `next` part: it potentially contains special attributes to be set with `setAttributeNS`.

Regarding the part to revert, I'm referring to [these lines](https://github.com/OlivierBlanvillain/monadic-html/blob/a1b7fa6564b5a38ddab753fb6e138dc7fe578033/monadic-html/src/main/scala/scala/xml/xml.scala#L26-L34) that I've added to merge namespaces with attributes, which is obviously not the correct way to do it :).

adrobisch:
Thanks for the clarification! Should not the merge still happen to have the `xmlns` attributes in the dom, regardless of where the prefixes are used in the end?

OlivierBlanvillain:
Oh I didn't think of that, I guess it's only a question of aesthetic? It's feels a bit weird to duplicate namespace information by putting it twice into the DOM...

adrobisch:
Yeah, its kind of duplicated, but personally I would prefer the rendered html to be as close to the original as possible (if only for debugging reasons), so I kept the merge. I added now the necessary methods (from my pov) to get and set the namespace during the mount.

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -9,21 +9,27 @@ import language.higherKinds
 // directly in scalac paser). `NoteSeq` disappeared, `Node <:!< Seq[Node]`...
 
 /** Trait representing XML tree. */
-sealed trait Node
+sealed trait Node {
+  def scope: Option[Scope] = None
+  def prefix: Option[String] = None
+  def namespace: Option[String] = if (prefix.isDefined) prefix.flatMap(namespaceURI) else scope.flatMap(_.namespaceURI(null))
```
OlivierBlanvillain: Could be simplified to `scope.flatMap(_.namespaceURI(prefix.orNull))`

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -91,6 +99,8 @@ sealed trait MetaData extends Iterable[MetaData] {
 
   def iterator: Iterator[MetaData] =
     Iterator.single(this) ++ next.iterator
+
+  def namespace: Option[String] = value.namespace
```
OlivierBlanvillain: I think this is not used anywhere

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -9,21 +9,27 @@ import language.higherKinds
 // directly in scalac paser). `NoteSeq` disappeared, `Node <:!< Seq[Node]`...
 
 /** Trait representing XML tree. */
-sealed trait Node
+sealed trait Node {
+  def scope: Option[Scope] = None
+  def prefix: Option[String] = None
+  def namespace: Option[String] = if (prefix.isDefined) prefix.flatMap(namespaceURI) else scope.flatMap(_.namespaceURI(null))
+  def namespaceURI(prefix: String): Option[String] = scope.flatMap(_.namespaceURI(prefix))
```
OlivierBlanvillain: I think this could be removed after https://github.com/OlivierBlanvillain/monadic-html/pull/36/commits/f5421a31f65dc112da95f9a12a10c43aa5db1617#r98707126

OlivierBlanvillain:
LGTM, many thanks, merging!

OlivierBlanvillain:
Released as `0.2.2`

PR #37, opened 2017-02-02
Add srcset as attribute

olafurpg:
See http://w3c.github.io/html/semantics-embedded-content.html#element-attrdef-img-srcset

### DIFF
--- a/monadic-html/src/main/scala/mhtml/settings.scala
+++ b/monadic-html/src/main/scala/mhtml/settings.scala
@@ -128,7 +128,7 @@ trait DevSettings extends MountSettings {
     "span"            , "spellcheck"      , "src"             , "srcdoc"          , "srclang"         ,
     "start"           , "step"            , "style"           , "tabindex"        , "target"          ,
     "title"           , "translate"       , "type"            , "usemap"          , "value"           ,
-    "width"           , "wrap"
+    "width"           , "wrap"            , "srcset"
   )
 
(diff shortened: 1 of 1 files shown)


olafurpg:
I don't know where to find an up-to-date list of legal attributes, would be nice to solve this once and for all instead of iteratively add one-by-one.

olafurpg:
Realized the list needs to be ordered for binary search to work. It would be nice if it was easier to add custom attributes to DevSettings.

OlivierBlanvillain:
```scala
trait DevSettings2 extends DevSettings {
  override lazy val attributes: Array[String] = ("yolo" :+ super.attributes).sorted
}
```

If we can find a good source for elements / attributes / events the best would be to have a curl + grep script to update these things automatically. Hardest part is probably aligning comas :P

OlivierBlanvillain:
The above does not quite work because of the `lazy val`, but making it a `def` should be enought

olafurpg:
That's what I'm doing, but I only want to use this for dev settings, so I ended up with

```scala
trait DevSettings2 extends DevSettings {
  lazy val superAttrs = new DevSettings{}.attributes
  override lazy val attributes: Array[String] = ("yolo" :+ superAttrs).sorted
}
if (LinkingMode.productionMode) MountSettings.default
else new DevSettings2
```

It should be fairly simple to wrap this into a helper method, something like `MountSetttings.custom(extraAttributes = ..., ...)`

PR #41, opened 2017-03-21
minor doc improvements

bbarker:
Fixing typo and adding `product` to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ This section presents the `Rx` API in its entirety. Let's start with the referen
 
     Dynamically switch between different `Rx`s according to the given
     function, applied on each element of this `Rx`. Each switch will cancel
-    the subscriptions for the previous outgoing `Rx`, and launch subscriptions
+    the subscriptions for the previous outgoing `Rx`, and launch the same subscriptions
     on the next `Rx`.
 
     Together with `Rx#map` and `Rx.apply`, flatMap, `Rx` is a `Monad`. [Proof](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-rx-cats/src/main/scala/mhtml/cats.scala).
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ This section presents the `Rx` API in its entirety. Let's start with the referen
 
     Dynamically switch between different `Rx`s according to the given
     function, applied on each element of this `Rx`. Each switch will cancel
-    the subscriptions for the previous outgoing `Rx`, and lunch a subscriptions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,7 +159,7 @@ This section presents the `Rx` API in its entirety. Let's start with the referen
 
     Dynamically switch between different `Rx`s according to the given
     function, applied on each element of this `Rx`. Each switch will cancel
-    the subscriptions for the previous outgoing `Rx`, and lunch a subscriptions
+    the subscriptions for the previous outgoing `Rx`, and launch the same subscriptions
```
OlivierBlanvillain: Maybe it should be `and lunch a new subscription on the next Rx`. It's not the same, it's really a new one!

For instance

```scala
rx1.flatMap(_ => rx2.fold(0)(_ + _))
```

Will systematically "reset" the `fold` on every `rx1` update.

README.md
```diff
@@ -191,6 +191,11 @@ This section presents the `Rx` API in its entirety. Let's start with the referen
     ```
 
     With this operation, `Rx` is a `Semigroup`. [Proof](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-rx-cats/src/main/scala/mhtml/cats.scala).
+    
+        
+- `def product[B](other: Rx[B]): Rx[(A, B)]`
+
+    Create the Cartesian product of two `Rx`; in other words, create an `Rx` of the tuple of the values in the source `Rx` (type `A`) and other `Rx` (type `B`). You can combine multiple `Rx` more succinctly using `|@|` syntactic sugar [as in Cats](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/typeclasses/applicative.md).
```
OlivierBlanvillain: This feels a bit imprecise, it's hard to tell from the text if this that this not actually a `zip`. 

Here is an example of how it should behave:

```scala
// r1      => 0     8                       9     ...
// r2      => 1           4     5     6           ...
// product => (0,1) (8,1) (8,4) (8,5) (8,6) (9,6) ...
```

(Here is how `zip` would behave instead, with the same signature:)

```scala
// r1  => 0     8                       9     ...
// r2  => 1           4     5     6           ...
// zip => (0,1)       (8,4)             (9,5) ...
```

To be consistent with the `flatMap` / `merge` doc, I would conclude with something like this

```
This method together with `Rx.apply` forms am `Applicative`. `|@|` syntax is available via the `monadic-rx-cats` package.
```

bbarker:
Thanks for the feedback! will try to fix soon

OlivierBlanvillain:
@bbarker I updated the doc in #43, if you want to give feedback there :)

PR #42, opened 2017-03-22
Add syntax to convert Future to Rx

OlivierBlanvillain:


OlivierBlanvillain:
@olafurpg That's would create an asymmetry with a potential sub project for `monix.Task`. I clearly don't want a monix dep on the main project, but I also don't want to make it more convenient to use `scala.concurrent.Future` over `monix.Task` since I strongly believe the latter is superior.

olafurpg:
The point of the `toRx` helper is to get rid of `scala.concurrent.Future`, the most common use case being to wrap `Ajax.{get,post}` requests like here: https://github.com/OlivierBlanvillain/monadic-html/blob/254e5cf2bb5115823b822a97e5abea34b92c03d3/examples/src/main/scala/mhtml/examples/GithubAvatar.scala#L46 

Is there a similar http client like `Ajax.get` based on monix?

bbarker:
Yes, RosHttp

olafurpg:
RosHttp is not free from scala.concurrent.Future https://github.com/hmil/RosHTTP/search?utf8=%E2%9C%93&q=scala.concurrent.Future&type= My point is just that monix and concurrent.Future are not mutally exclusive. I don't see the need for a `-future` specific module if you can't avoid scala.concurrent.Future in the first place. If there exists a scala.concurrent.Future free http client, then I'm 👍 for creating a `-future` module.

OlivierBlanvillain:
`Ajax.{get,post}` is [library based](https://github.com/scala-js/scala-js-dom/blob/master/src/main/scala/org/scalajs/dom/ext/Extensions.scala#L301-L326), the fact that it's published by default with `scala-js-dom` is not a sufficient argument to give a privileged status to `scala.concurrent.Future`.

Also a port to monix Task should be trivial: https://gist.github.com/OlivierBlanvillain/4105d05aea258b56f9317b3c81072118

bbarker:
"I also don't want to make it more convenient to use scala.concurrent.Future over monix.Task since I strongly believe the latter is superior." 

Future seems to be more frequent in occurrence than monix, partly because it is standard (not because it is better). I feel like this comes up frequently enough that it is worth merging. A FAQ or gist could be linked to if we want to make it easy for people to add something similar when using monix, and if nothing else, the Future example provides a good template for people to start from when implementing conversions from other async libraries.

OlivierBlanvillain:
Alright, I'm sold, I'm going to get back to this PR and merge something for `scala.concurrent.Future`

PR #43, opened 2017-03-23
Remove .impure.value and document product

OlivierBlanvillain:
`.impure.value` creates confusion: it makes it look like `Rx` have a current value when they actually don't.

This can be particularly surprising with expressions like `rx.fold(0)(_ + _).impure.value` that will always evaluate to `0` because `.value` creates and cancels its subscription on every call...

So my proposal it to remove this altogether.

I also created an `RxImpureOps` to replace the `val impure = object {}`. This should make it easier to add impure methods like `.impure.oreachNext` & co.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,10 +159,26 @@ This section presents the `Rx` API in its entirety. Let's start with the referen
 
     Dynamically switch between different `Rx`s according to the given
     function, applied on each element of this `Rx`. Each switch will cancel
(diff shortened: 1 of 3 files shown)


bbarker:
Looks good to me; minor spelling typo: lunch -> launch

OlivierBlanvillain:
Thanks!

PR #44, opened 2017-03-25
doc examples in FAQ for HTML entities and mounting Node[Seq]

bbarker:


README.md
```diff
@@ -319,6 +319,35 @@ Elements:
...
+}
+```
+
+#### How do I use HTML entities?
+
+There are a couple of ways. The first is to inject it directly into a node. To get two non-breaking spaces, you might do:
+`<span>&npsp;&npsp;</span>`. More generally, you will need `scala.xml.EntityRef` to reference an entity directly in Scala:
+
+```scala
+def space(nn: Int): Node = Group((0 until nn).map{_ => EntityRef("nbsp")})
```
OlivierBlanvillain:
```scala
Seq.fill(nn)(EntityRef("nbsp"))
```

README.md
```diff
@@ -319,6 +319,35 @@ Elements:
...
+<span>{space(2)}</span> // results in  <span>"&npsp;""&npsp;"</span>
+
```
OlivierBlanvillain: Extra new line

README.md
```diff
@@ -319,6 +319,35 @@ Elements:
...
+```
+
+#### How do I use HTML entities?
+
+There are a couple of ways. The first is to inject it directly into a node. To get two non-breaking spaces, you might do:
+`<span>&npsp;&npsp;</span>`. More generally, you will need `scala.xml.EntityRef` to reference an entity directly in Scala:
+
+```scala
+def space(nn: Int): Node = Group((0 until nn).map{_ => EntityRef("nbsp")})
+<span>{space(2)}</span> // results in  <span>"&npsp;""&npsp;"</span>
```
OlivierBlanvillain: Why are there quotes (`"`) here?

bbarker: That's how it shows up in the DOM inspector; here's an example from `<h1 style = "color: #FFFFFF">{space(3)}{ced2ar}</h1>`:
image

bbarker: Maybe a scala.xml thing?

bbarker:
Added the test, good way to get warmed up with testing mhtml!

README.md
OlivierBlanvillain: Hmm that's weird, but whatever if the browser renders things expected.

OlivierBlanvillain:
Many thanks, merging :-)

PR #49, opened 2017-04-04
adding initial toVar implementation

bbarker:
This is a first pass. Might want to try more tests, but I couldn't think of much today. Then, I'll try to test it in TodoMVC in the next RC/Snapshot.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.3.0"
+version in ThisBuild := "0.3.1-SNAPSHOT"


monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -154,6 +154,13 @@ case class RxImpureOps[+A](self: Rx[A]) extends AnyVal {
    * Omitting to canceling subscription can lead to memory leaks.
    */
   def foreach(effect: A => Unit): Cancelable = Rx.run(self)(effect)
+
+  /**
+    * Returns a Var with the same subscriptions as the original Rx.
+    * Useful for creating a Var from an Rx that is not a Var.
+    */
+  def toVar[B >: A]: Var[B] = new Var[B](None, self => foreach(self.:=))
```
OlivierBlanvillain: `name` shadows the argument of `RxImpureOps`

bbarker:
I believe I fixed the shadowing issue; odd, though - IntelliJ usually catches those, but IntelliJ seems to have some other issues in rx.scala that may be the root cause, but I haven't looked into it.

I added a variation of your tests, and added a commented out version based on Merge, but I may not be using the internal API correctly. I do now take your point about the possible issue in the current implementation of order not being well defined, which I tried to address in the not-yet-working-correctly merge-based version.

PR #52, opened 2017-04-21
Fixed broken links in README

Kemichal:
Fixed broken links and changed some to be relative.


OlivierBlanvillain:
I think I would prefer to only have absolute links, the reason being that CI checks them using [this line](https://github.com/OlivierBlanvillain/monadic-html/blob/master/.travis.yml#L15).

Would you mind updating your PR? Thanks!

OlivierBlanvillain:
LGTM, many thanks!

PR #54, opened 2017-04-22
Added copy method to Elem

Kemichal:
Even though `Elem` is a case class it doesn't have a copy method because it has a [repeated parameter](https://www.scala-lang.org/files/archive/spec/2.12/05-classes-and-objects.html#case-classes).

### DIFF
--- a/monadic-html/src/main/scala/scala/xml/xml.scala
+++ b/monadic-html/src/main/scala/scala/xml/xml.scala
@@ -47,7 +47,7 @@ final case class Elem(
     label: String = this.label,
     attributes1: MetaData = this.attributes1,
     scope: Option[Scope] = this.scope,
-    child: Seq[Node] = this.child.toSeq): Elem = Elem(prefix, label, attributes1, scope, child: _*)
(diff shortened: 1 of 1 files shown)


monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -41,6 +41,13 @@ final case class Elem(
       }
       merge(a, s)
     }, Some(s), c: _*)
+
+  def copy(
+    prefix: Option[String] = this.prefix,
+    label: String = this.label,
+    attributes1: MetaData = this.attributes1,
+    scope: Option[Scope] = this.scope,
+    child: Seq[Node] = this.child.toSeq): Elem = Elem(prefix, label, attributes1, scope, child: _*)
```
OlivierBlanvillain: Why `.toSeq`?

Kemichal: Yea, it doesn't seem to be needed, fixed in new commit.

OlivierBlanvillain:
Thanks! :+1:

PR #58, opened 2017-05-08
Add isHot method

OlivierBlanvillain:
[See discussion on gitter](https://gitter.im/monadic-html/Lobby?at=59109bfb33e9ee771c873f1d). @fizzy33 want to review?


fizzy33:
+1 from me.  The other thing that is handy would be a way to get the # of subscribers.  We would use that on our long lived apps to do some sanity checks.  It doesn't need to be exposed directly.  Like a 
def subscriberCount[A](rx: Rx[A]): Int 
Hidden in like an MhtmlInternals objects.  So it wouldn't be part of the public api but can be used for metering the app.

OlivierBlanvillain:
In scala this is basically what package private mean, you can hijack them from anywhere:

```scala
package my.org
package mhtml

object Internal {
  implicit class subscriberCountOp[T](rx: Rx[T]) {
    def subscriberCount[A](rx: Rx[A]): In = rx.subscribers.size
  }
}
```

fizzy33:
Good point and that works for me...

PR #53, opened 2017-04-21
adding compiling but runtime-failing Todo-MVC example (WIP)

bbarker:
A few quick notes:
1. I agree this probably isn't the best example to demonstrate components ... except that it is a sort of canonical example, and if it components can be done with TodoMVC, then maybe it should be, since it gives a quick way to demonstrate how to build a somewhat larger SPA (just my thoughts on it - maybe it is infeasible).
2. I basically demonstrated what you said about having implemented fold by using lazy vals ... by explicitly using `foldp` in the current definition of `allTodos` (old definition is commented out). Though I have to admit, I'm not sure if this is correct since it is wrapped inside of a `flatMap` call.
3. I'm getting a MatchError inside of `rx.scala` with both implementations of `allTodos` ... any insights on how to debug this sort of error would be appreciated as well :-) : 
image

PR #59, opened 2017-05-09
Fix xml to string

OlivierBlanvillain:
The cause of the problem was that `MetaData` where getting an unsound default `toString` implementation from extending `Iterable`... On fix is to explicitly override `toString` in `MetaData` (or its sub classes), the other, much cleaner, is to remove this `MetaData extends Iterable[MetaData]` madness.

@bbarker quick review? This is a clean fix for the `override def toString` I've pushed in #53...


bbarker:
LGTM! I've run into the `extends Iterable[MetaData]` weirdness before as well I believe. Pretty soon, this xml impl will be popular enough we may need a scala xml typeclass library ;).

PR #60, opened 2017-05-14
initial call location counting for Vars

bbarker:
This seems to work ... except in one case. For some reason, the doge example looks as though it doesn't need allowMultiMutations but I had to set `val count: Var[Int] = Var(0, allowMultiMutations = true)` for it to work. 

Currently the option is not configurable on the JVM; non-production is assumed in that case.

### DIFF
--- a/monadic-rx/shared/src/main/scala/mhtml/rx.scala
+++ b/monadic-rx/shared/src/main/scala/mhtml/rx.scala
@@ -241,8 +241,8 @@ object Rx {
 
 /** A smart variable that can be set manually. */
 class Var[A](
-  initialValue: Option[A], register: Var[A] => Cancelable,
-  allowMultiMutations: Boolean = false
+  initialValue: Option[A],
+  register: Var[A] => Cancelable, allowMultiMutations: Boolean = false
 ) extends Rx[A] {
   // Last computed value, retained to be sent to new subscribers as they come in.
   private[mhtml] var cacheElem: Option[A] = initialValue
@@ -250,23 +250,33 @@ class Var[A](
   private[mhtml] var registration: Cancelable = Cancelable.empty
   // Mutable set of all currently subscribed functions, implementing with an `Array`.
   private[mhtml] val subscribers = buffer.empty[A => Unit]
+
   // Keeps track of mutable call locations
-  private[mhtml] var mutationLocations: Set[Int] = Set()
+  final case class MutationInfo(file: String, line: Int, hash: Int)
+  private[mhtml] var mutationLocations: List[MutationInfo] = List()
   private[mhtml] def wasSet: Boolean = mutationLocations.size > 1
 
-  private[mhtml] def addUpdateCall(): Unit = {
-    if (!Var.isProduction) {
-      if (!allowMultiMutations) {
-        // Avoid doing expensive safety checks in optimized code that
-        // should be caught during testing.
-        val straceSeq = new RuntimeException().getStackTrace.toSeq
-        val lineNumHash = straceSeq.map(stl => stl.getLineNumber).mkString("\n").hashCode.toString
-        val contentHash = straceSeq.map(
-          stl => stl.getFileName + stl.getClassName + stl.getClassName
-        ).mkString("\n").hashCode.toString
-        val hashCode: Int = (lineNumHash + contentHash).hashCode
-        mutationLocations += hashCode
-      }
+  private[mhtml] def addUpdateCall(updateType: String): Unit = {
+    if (!Var.isProduction && !allowMultiMutations) {
+      // Avoid doing expensive safety checks in optimized code that
+      // should be caught during testing.
+      val straceSeq = new RuntimeException().getStackTrace.toSeq
+      val lineNumHash = straceSeq.map(stl => stl.getLineNumber).mkString("\n").hashCode.toString
+      val contentHash = straceSeq.map(
+        stl => stl.getFileName + stl.getClassName + stl.getClassName
+      ).mkString("\n").hashCode.toString
+      val hashCode: Int = (lineNumHash + contentHash).hashCode
+      mutationLocations = MutationInfo(
+        straceSeq(2).getFileName, straceSeq(2).getLineNumber, hashCode
+      ) :: mutationLocations
+    }
+    if (wasSet) {
+      val mut2 = mutationLocations.head
+      val mut1 = mutationLocations(1)
+      throw new IllegalStateException(
+        s"$updateType attempted at [${mut2.file} (${mut2.line})];" +
+          s" but Var was already set at [${mut1.file} (${mut1.line})];"
+      )
     }
   }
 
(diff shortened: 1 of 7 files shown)


monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -268,6 +291,13 @@ class Var[A](initialValue: Option[A], register: Var[A] => Cancelable) extends Rx
 
   /** Sets the value of this `Var`. Triggers recalculation of depending `Rx`s. */
   def :=(newValue: A): Unit = {
+    addUpdateCall()
+    if (wasSet) {
```
OlivierBlanvillain: I would put all this if block inside `addUpdateCall`

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -306,8 +341,12 @@ object Var {
    */
   def create[A](initialValue: A)(register: Var[A] => Cancelable): Var[A] =
     new Var[A](Some(initialValue), register)
+
+
+  private[mhtml] lazy val isProduction: Boolean = Platform.checkIsProduction
```
OlivierBlanvillain: I would guess going through a member (and a lazy val?) might prevent stuff from being constant folded (see [this scala doc](https://github.com/scala-js/scala-js/blob/master/library/src/main/scala/scala/scalajs/LinkingInfo.scala#L16-L40)). We need check the emitted IR to see if the `isProduction` branch are properly DCE under `fullOptJS`...

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -237,13 +237,36 @@ object Rx {
...
+  allowMultiMutations: Boolean = false
+) extends Rx[A] {
   // Last computed value, retained to be sent to new subscribers as they come in.
   private[mhtml] var cacheElem: Option[A] = initialValue
   // Current registration to the feeding `Rx`, canceled whenever nobody's listening.
   private[mhtml] var registration: Cancelable = Cancelable.empty
   // Mutable set of all currently subscribed functions, implementing with an `Array`.
   private[mhtml] val subscribers = buffer.empty[A => Unit]
+  // Keeps track of mutable call locations
+  private[mhtml] var mutationLocations: Set[Int] = Set()
```
OlivierBlanvillain: I would use the same mutable buffer used for `subscribers`, that might save putting all of `scala.collection.immutable.Set` in the js. Maybe it could be initialized at `null` to reduce the memory footprint in production?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -237,13 +237,36 @@ object Rx {
...
+  private[mhtml] def addUpdateCall(): Unit = {
+    if (!Var.isProduction) {
+      if (!allowMultiMutations) {
+        // Avoid doing expensive safety checks in optimized code that
+        // should be caught during testing.
+        val straceSeq = new RuntimeException().getStackTrace.toSeq
+        val lineNumHash = straceSeq.map(stl => stl.getLineNumber).mkString("\n").hashCode.toString
+        val contentHash = straceSeq.map(
+          stl => stl.getFileName + stl.getClassName + stl.getClassName
+        ).mkString("\n").hashCode.toString
```
OlivierBlanvillain: It might be interesting to see if we could keep more info about this stack-trace to say something like "error: var was mutate at two locations, the first time from <file:line>, the second time from <file:line>.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -293,6 +326,8 @@ object Var {
   /** Create a `Var` from an initial value. */
   def apply[A](initialValue: A): Var[A] =
     new Var[A](Some(initialValue), _ => Cancelable.empty)
+  def apply[A](initialValue: A, allowMultiMutations: Boolean): Var[A] =
```
OlivierBlanvillain: Any reason not to use a default argument here?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I was thinking the same thing, I will try it in the next round. Definitely worthwhile to have.

bbarker:
I like the idea of having multiple apply methods, but I would argue slightly in favor of having `apply` be the safer option by default. Although they might be surprised the first time they see an error, that surprise will be quickly overcome, whereas the surprise from mutating in multiple locations might be worse and be less easy to overcome.  "Safe by default, extra syntax when wanting to be unsafe" seems like a generally good principle to me.

bbarker:
Since this code will be optimized away in production, I could even have the error message direct them to the correct apply method should they want to use multiple mutations. Maybe 'Var.unlimited`.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -237,13 +237,36 @@ object Rx {
 }
 
 /** A smart variable that can be set manually. */
-class Var[A](initialValue: Option[A], register: Var[A] => Cancelable) extends Rx[A] {
+class Var[A](
+  initialValue: Option[A], register: Var[A] => Cancelable,
```
bbarker: I have to admit I haven't been using scalafmt on any project so far; is that what I should be using?

OlivierBlanvillain: ahah I don't know I just do everything by hand :P

OlivierBlanvillain:
> Although they might be surprised the first time they see an error, that surprise will be quickly overcome, whereas the surprise from mutating in multiple locations might be worse and be less easy to overcome.

Then it must definitely be an warning instead of an exception, runtime errors are just too unfriendly in my opinion.

The other point that make me make slightly hesitant is that I don't want to be too opinionated without having a super clean/simpler alternative to propose. Maybe after hacking some more on the TodoMVC we realise that a single multi location mutated `Var` for all events is actually a reasonable pattern...

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I'll give this a try - any chance of merging #64 first?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: How about instead of using a simple Boolean, we use an implicit parameter that wraps a Boolean? - then the user can specify how they want their vars to be handled by default, and not rely on a default defined by the mhtml project.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I searched for strings appearing in the `IllegalStateException` thrown from `addUpdateCall` inside of `monadic-rx/js/target/scala-2.12/monadic-rx-opt.js` and didn't see anything (compared to `monadic-rx/js/target/scala-2.12/monadic-rx-test-fastopt.js`, where I could easily see the entire function definition for `addUpdateCall`).

OlivierBlanvillain:
Changes look fine to me! I think I would still prefer a warning instead of an exception, as runtime exceptions make for a very bad first time interaction with a library.

Also I propose that we give the `imitate` construct a shot before deprecating multi spot mutations on a var :)

PR #63, opened 2017-05-19
Rename product to zip

OlivierBlanvillain:
The `product` name makes more sense from a cats point of view, but cats users are probably going to use the raisin bread (|@|) operator regardless. `zip` is the other commonly used name (standard library, scalaz) that might be the first coming to the mind of most scala users.

### DIFF
--- a/monadic-rx/shared/src/main/scala/mhtml/rx.scala
+++ b/monadic-rx/shared/src/main/scala/mhtml/rx.scala
@@ -33,15 +33,18 @@ sealed trait Rx[+A] { self =>
    * `for { a <- ra; b <- rb } yield (a, b)`.
    *
    * ```
-   * // r1      => 0     8                       9     ...
-   * // r2      => 1           4     5     6           ...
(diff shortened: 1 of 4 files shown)


monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -33,15 +33,18 @@ sealed trait Rx[+A] { self =>
...
    * ```
    *
    * This method, together with `Rx.apply`, forms am `Applicative`.
    * `|@|` syntax is available via the `monadic-rx-cats` package.
    */
-  def product[B](other: Rx[B]): Rx[(A, B)] = Product[A, B](this, other)
+  def zip[B](other: Rx[B]): Rx[(A, B)] = Zip[A, B](this, other)
+
+  @deprecated("Renamed to zip", "0.3.3")
+  def product[B](other: Rx[B]): Rx[(A, B)] = zip(other)
```
olafurpg: It should be easy to provide a scalafix rewrite for this migration. If you're interested

PR #64, opened 2017-07-27
making use of scala.js collection API for buffer

bbarker:
This consolidates buffer code into the shared part of the project and makes typing of subscriber arrays much easier to deal with. I'm not sure when this change happened in scala.js to make this possible, but the release notes for 0.6.0 seem relevant.


bbarker:
This also makes looking at rx.scala more pleasant in IntelliJ (I know, don't use IntelliJ ... ;-)).

OlivierBlanvillain:
`collection.mutable.ArrayBuffer` is significantly slower than native array on js... Do you need anything in particular from `ArrayBuffer` that you can't get from native arrays?

bbarker:
I don't think we need anything from ArrayBuffer specifically, I just stumbled upon it as a common type that could be used for both JS and the JVM. I don't think I tried, but can we use `Array` directly instead?

OlivierBlanvillain:
IIRC `Array`s are not resizable, whereas native js arrays are

bbarker:
@OlivierBlanvillain Some possible solutions are discussed at https://stackoverflow.com/questions/45425763/is-there-a-common-interface-for-js-array-in-scala-js-and-scala-on-the-jvm

OlivierBlanvillain:
Tests are optimised down to 630K on master:

    $ git checkout master
    $ sbt clean monadic-rxJS/test:fullOptJS
    $ ls -lAh monadic-html/monadic-rx/js/target/scala-2.12/monadic-rx-test-opt.js
    -rw-r--r-- 1 o o 630K Aug  1 14:36 monadic-rx-test-opt.js

636K on this branch:

    $ git checkout 1b3f1e2f
    $ sbt clean monadic-rxJS/test:fullOptJS
    $ ls -lAh monadic-html/monadic-rx/js/target/scala-2.12/monadic-rx-test-opt.js
    -rw-r--r-- 1 o o 636K Aug  1 14:38 monadic-rx-test-opt.js

So that's either 6K of js dead code, or a slower resulting binary...

bbarker:
I pushed a version using `ArrayBuilder`; unfortunately `ArrayBuilder` does not have `-=`, so it seems necessary to construct a new `ArrayBuilder` whenver that is used, which also means we need a `var` instead of a `val` for `private[mhtml] var subscribers:`. I suppose this could be avoided by making our own version of ArrayBuilder, but that would seem like overkill, especially considering cross-platform support. Maybe better to consider an appeal (if it makes sense) to the next generation collection maintainers.


The good news: there is now only 1K of difference between the master and this branch:

```
brandon@brandon-750-170se:~/workspace/monadic-html$ ls -lAh monadic-rx/js/target/scala-2.12/monadic-rx-test-opt.js
-rwxrwxrwx 1 root root 634K Aug 20 11:11 monadic-rx/js/target/scala-2.12/monadic-rx-test-opt.js
brandon@brandon-750-170se:~/workspace/monadic-html$ ls -lAh monadic-rx/js/target/scala-2.12/monadic-rx-test-opt.js
-rwxrwxrwx 1 root root 633K Aug 20 11:16 monadic-rx/js/target/scala-2.12/monadic-rx-test-opt.js
```
(Although not obvious from the paste, the first `ls` was for this branch, the second was for master).

OlivierBlanvillain:
It's still not very clear to me how much we would gain with this compared to the current approach. I'm more leaning toward keep using js array on js, and implemented whatever we need on top of that...

bbarker:
> and implemented whatever we need on top of that
IIRC, that's sort of what ArrayBuilder is doing, except not really *everything* as we saw with `-=`.

With the current approach, it seems types for buffer arrays cannot be resolved in IDEs. I believe I also ran into a more serious typing issue that was not IDE related when trying out some things in `rx.scala` but I'd have to dig about and find the branch and test it out - I think it was in my attempt at `imitate`.

bbarker:
Here is an example where needing a result type is important:

```scala
  private[mhtml] def getSubscribers[A](rxIn: Rx[A]) = rxIn match {
    case rxv: Var[A] => rxv.subscribers
    case rxp: Map[A, _] => getSubscribers(rxp.self)
    case rxp: FlatMap[A, _] =>  getSubscribers(rxp.self)
    case rxp: Zip[A, _] => getSubscribers(rxp.self)
    case rxp: DropRep[A] =>  getSubscribers(rxp.self)
    case rxp: Merge[A, A] => getSubscribers(rxp.self) ++ getSubscribers(rxp.other)
   // ...
  }
```

This will give the error `recursive method getSubscribers needs result type`; I don't think this can be fixed without having a common type on JS and JVM.

OlivierBlanvillain:
What about

```scala
package mhtml   
    
object buffer {    
  type Buffer[E] = Array[E]
  def empty[A] = ...
  def apply[A](size: Int) = ...
}
```

bbarker:
LGTM, though IntelliJ still has import issues with importing buffer.Buffer from rx.scala. I'll start a separate PR (#71) and close this one.

PR #66, opened 2017-08-02
Add imitate construct

OlivierBlanvillain:
This PR adds the `imitate` construct, inspired from [xstream's imitate](https://github.com/staltz/xstream/blob/7a6f69e0834bd23a511333bb1d0cf6cf923b400f/src/index.ts#L1845-L1909) (thanks @bbarker for the idea!).

This implementation doesn't require an additional impure method but adds an extra boolean field to all `Var`. An alternative would be to add a new subclass of `Rx`, `ImitatingVar`, without `:=` but with `imitate` instead. This would increase the surface API to lower the memory footprint, probably not worth doing.

Best reviewed commit by commit :smile:


bbarker:
LGTM. Tested it in a TodoMVC app with cycles.

OlivierBlanvillain:
Great, could you update #53 with whatever you have on the TodoMVC? I'm going to merge this PR for now and but wait for an proper example before making another release.

OlivierBlanvillain:
I just realised I completely disregarded the case where someone tries to `imitate` a `Var` twice. The current implementation will do first one and silently ignore the second one. It would be nice to use the same mechanism than in #60 to add "don't to that" warning.

bbarker:
I didn't forget, but thought that could be done as a second PR. Will be
happy to get to it next week!

PR #68, opened 2017-08-17
fixes #67: support numeric EntityRefs

### LINKED ISSUE #67: EntityRef("#182")
`transformEntityRef` needs to be update to handle this syntax, expected behavior is: https://alvinalexander.com/scala/scala-xml-entityref-example-unicode-character

bbarker:


monadic-html/src/main/scala/mhtml/unicode.scala
```diff
@@ -0,0 +1,27 @@
+package mhtml
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation._
+
+private[mhtml] object unicode {
+
+  // Note: NativeJSString and fromCodePoint are replicas of private
+  // methods/objects in Scala.JS
```
OlivierBlanvillain: I wonder if there is hack to do this conversion without copy pasting...

OlivierBlanvillain:
I think you are doing more work than `scala-xml` here:

```scala
scala> import xml._
import xml._

scala> <div>{EntityRef("#182")}</div>
res1: scala.xml.Elem = <div>&#182;</div>

scala> <div>&#182;</div>
res2: scala.xml.Elem = <div>¶</div>
```

I think #67 might not actually be a real issue?

OlivierBlanvillain:
I think we are way overdoing these things, here are the only transformation on `scala-xml` EntityRef done by in `scala-xml`: https://github.com/scala/scala-xml/blob/642b853295794dadfef873a9ecbd30dba4af39f1/shared/src/main/scala/scala/xml/EntityRef.scala#L24-L30

OlivierBlanvillain:
Closing this given what we just discussed on gitter. Thanks @bbarker for the PR! Somehow I'm still understanding new things about scala.xml everytime I work on this project ^^'

PR #72, opened 2017-08-28
adding merge for list of Rx (MergeN)

bbarker:


monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
```diff
@@ -199,6 +199,32 @@ class RxTests extends FunSuite {
     assert(rx1.isCold && rx2.isCold)
   }
 
+  test("mergeN") {
+    // Should be identical to merge test above
+    val rx1: Var[Int] = Var(0)
+    val rx2: Var[Int] = Var(1)
+    val rxs = List(rx1, rx2)
+    val merged: Rx[Int] = rxs.merge
```
bbarker: It seems the code from this point in the test can easily be factored out, so assuming we want to do this, I will plan to use a common function from `test("merge")` and `test("mergeN")`

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -211,6 +217,11 @@ object Rx {
       val c2 = run(other)(effect)
       Cancelable { () => c1.cancel; c2.cancel }
 
+    case MergeN(self) => {
+      val ccs: List[Cancelable] = self.map(rx => run(rx)(effect))
```
bbarker: Should we use `List` or something further up the inheritance hierarchy (`Iterable` ?)

monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
bbarker: Also, would need to add at least some doc to the README.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -165,11 +166,16 @@ object Rx {
   final case class Zip     [A, B]     (self: Rx[A], other: Rx[B])                   extends Rx[(A, B)]
   final case class DropRep [A]        (self: Rx[A])                                 extends Rx[A]
   final case class Merge   [A, B >: A](self: Rx[A], other: Rx[B])                   extends Rx[B]
```
OlivierBlanvillain: What do you think about remove `Merge` from here and implement it using `MergeN`?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -165,11 +166,16 @@ object Rx {
...
+  final case class MergeN  [A, B >: A](self: List[Rx[A]])                           extends Rx[B]
```
OlivierBlanvillain: I would make it an explicit non empty list:

```scala
final case class class MergeN   [A, B >: A](self: Rx[A], other: List[Rx[A]]) extends Rx[B]
```

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -165,11 +166,16 @@ object Rx {
...
   final case class Foldp   [A, B]     (self: Rx[A], seed: B, step: (B, A) => B)     extends Rx[B]
   final case class Collect [A, B]     (self: Rx[A], f: PartialFunction[A, B], b: B) extends Rx[B]
   final case class SampleOn[A, B]     (self: Rx[A], other: Rx[B])                   extends Rx[A]
   final case class Imitate [A]        (self: Var[A], other: Rx[A])                  extends Rx[A]
 
+  implicit class RichListRx[A](val value: List[Rx[A]]) extends AnyVal {
```
OlivierBlanvillain: Not sure about adding stuff to `List`... I would simply add that to the other methods with a varargs, it's would also be better for discoverability:

```scala
rx.mergeN(rxb, rxc, rxd)
```

monadic-rx/shared/src/main/scala/mhtml/rx.scala
OlivierBlanvillain: I guess if you go for varargs this is going to change to `Seq` (either way works for me)

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I don't know the exact performance implications, but since merging two Rx's is probably the most common situation in practice, it might be good to leave it in.

monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
OlivierBlanvillain: I wouldn't factor out the tests. I think it a nice feature to have each test as a small isolated example, they are simpler to read and write :) (and should have zero maintenance cost, assuming backwards compatibility)

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: But what is `self` supposed to be here?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: Having it on list feels more natural, once you know about it, and it should only work on `List[Rx[...]]` - at least, IntelliJ seems smart enough to get the completion correct only in the valid case (for what it's worth). We could also provide both interfaces.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
OlivierBlanvillain:
```
rxa.mergeN(rxb, rxc, rxd)
↑ this guy, "rxa"
```

monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
bbarker: OK. I guess I just feel that code duplication, even in tests, can be tricky. If a person is looking at one test, but forgets another test that is supposed to be the same, then over time they might begin to evolve into different organisms ;).

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I'm hesitant to explicitly require `Seq`, it seems particularly bad:

https://danielasfregola.com/2015/06/15/which-immutable-scala-collection/

https://youtu.be/DXFUn90hudk?t=35m30s

Maybe we shouldn't fret too much about performance at this point, bench marks in use cases can be done later, as we aren't at 1.0 yet.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
OlivierBlanvillain: Your version would fail at runtime on `List.empty[Rx[Uni]].merge`. Also, the clean way to add methods on standard types would be do to it under an `mhtml.syntax._` import, which definitely feels overkill for a single method...

Another way would be `Rx.mergeN(rxa, rxb, rxb)`.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I thought so, but that approach looks a bit awkward to me at first glance. It will typically require the construction of some dummy Rx for `rxa`.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: Hmm, this did not generate any runtime errors for me, but maybe I misunderstood:

```scala
  test("RichListRx") {
    val x = List.empty[Rx[Unit]].merge
  }
```


There has to be a first method ;) More seriously though, I suspect there will be others to follow as we flesh out what we want to do on lists of Rx in practice.

I do like the `Rx.mergeN` option better than `rxa.mergeN(rxb, rxc, rxd)`, but would still be nice if the optional `mhtml.syntax` import as you suggest. I guess it is entirely possible for the user to create their own `RichListRx` since `MergeN` is not `private`, but I don't see the harm in supplying a few for convenience and examples; we could even potentially have alternatives for different collections (`List`, `Seq`, `Array`, ..) that might be too much, just an idea.

OlivierBlanvillain:
```scala
val x = List.empty[Rx[Unit]].merge
```

This is an empty Rx, which breaks the fundamental non empty assumption of every Rx in mhtml :P 

Let's keep skip the syntax for now, a simple API like this should do the trick:

```scala
object Rx {
  def mergeN[Rx[A]](head: Rx[A], tail: Rx[A]*): Rx[A]
}
```

bbarker: Oh, I see, that would break things.

Here's another solution I think: 
```scala
    case MergeN(self) =>
      val selfRxs: List[Rx[A]] = self.headOption match {
        case Some(_) => self
        case None => List(Rx(null.asInstanceOf[A]))
      }
      val ccs: List[Cancelable] = selfRxs.map(rx => run(rx)(effect))
      Cancelable { () => ccs.foreach(cc => cc.cancel) }
```

Some minimal runtime overhead there for the check, but I guess there would also be some runtime overhead every time a user has to remove an element from the head of a list, the prepend it back (more or less depending on what collection we are using), with the additional typing/screenspace overhead of having to do so each time.

OlivierBlanvillain:
(Sorry for the long silence.)

I've just had a quick look in cats, and there are already a [`MergeN`](https://github.com/typelevel/cats/blob/7be0e6433ff8fe56bbe3bffc9b9048d0602e43e2/kernel/src/main/scala/cats/kernel/Monoid.scala#L27-L30) that works on any monoid:

```
val ra: Rx[Int] = ...
val rb: Rx[Int] = ...
Monoid.combineAll(List(ra, rb)): Rx[List[Int]]
```

It's not very clear to me if the (small?) performance gain we would get from a specialized operator for this case is worth the extra complexity, it could be premature optimization...

bbarker:
Cool, I'll try the Cats alternative

PR #74, opened 2017-09-07
Remove settings

OlivierBlanvillain:
This PR removes the mounting settings following up the discussion we had on #68.

To be followed by a separate PR to inject ` xml.Node => xml.Node` and `xml.MetaData => xml.MetaData` functions. (I tried the fix point approach, but it was too heavy weight)


bbarker:
So the xml Node and MetaData transformation functions will be a means to inject specific dev settings, for instance? That sounds like a simpler approach, so sounds good to me. 

Even though we have less code for this, it might not hurt to leave in some of the EntityRef tests.

LGTM, otherwise.

PR #79, opened 2017-10-22
working on a foldp test/demo for rt

bbarker:
I suspect a logic error on my part, not sure why though; even if `assert(a.foldp(0)(2*_ + 2*_).impure.value == 6)`, I would expect `assert(b.impure.value == 6)` to work.

### DIFF
--- a/monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
+++ b/monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
@@ -114,6 +114,26 @@ class RxTests extends FunSuite {
     assert(value.isCold)
   }
 
+  test("Referential transparency with foldp") {
+
+    val a: Var[Int] = Var(0)
+    val b: Rx[Int] = a.foldp(0)(2*_ + 2*_)
+    // Should be:  2*0 + 2*0 == 0
+    assert(b.impure.value == 0)
+    assert(a.foldp(0)(2*_ + 2*_).impure.value == 0)
+    a := 1
+    // Should be:  2*0 + 2*1 == 2
+    assert(b.impure.value == 2)
+    assert(a.foldp(0)(2*_ + 2*_).impure.value == 2)
+    a := 1
+    // Should be:  2*2 + 2*1 == 6
+    assert(b.impure.value == 6)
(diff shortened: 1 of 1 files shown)


monadic-rx/shared/src/test/scala/mhtml/RxTests.scala
```diff
@@ -76,6 +76,26 @@ class RxTests extends FunSuite {
...
+    val b: Rx[Int] = a.foldp(0)(2*_ + 2*_)
+    // Should be:  2*0 + 2*0 == 0
+    assert(b.impure.value == 0)
+    assert(a.foldp(0)(2*_ + 2*_).impure.value == 0)
+    a := 1
+    // Should be:  2*0 + 2*1 == 2
+    assert(b.impure.value == 2)
+    assert(a.foldp(0)(2*_ + 2*_).impure.value == 2)
+    a := 1
+    // Should be:  2*2 + 2*1 == 6
```
OlivierBlanvillain: That's not how it works, `Rx` do not remember all their history. When you interpret an expression containing a `foldp` its only evalued with the current values the `Rx` (and not using all history).

Here is what is written in the documentation of `foldp`:

> Produces a Rx containing cumulative results of applying a binary operator to each element of this Rx, starting from a seed and the current value, and moving forward in time.

So when you do `[...]; a := 1; a.foldp(0)(2*_ + 2*_).impure.value`, this *always* evaluates to `2`, no mater what you write in `[...]`.

bbarker: Ah, I see, right - "current value" in this case means the current value of `a`, which makes great sense to me from a non-stateful point of view, now. At the time, I was thinking "current value" was something internal to `foldp`.

OlivierBlanvillain: That's indeed an important detail :smile: Maybe could be made explicit in the docs: `and the current value of this Rx,`? The sentence becomes a bit cluttered but at least it's unambiguous.

bbarker:
@OlivierBlanvillain finally updated this one

PR #82, opened 2017-10-31
[WIP] add Rx.sharing 

bbarker:
I like both suggestions mentioned for sharing in #70 , but since the second largely depends on the first, I thought I'd try to get the ball rolling with the first. Unfortunately, updating upstream var doesn't propagate. When `run` operates on `Sharing`, the `foreach` (` self.impure.foreach{ x => ...`) doesn't seem to execute for updated values of the upstream Rx, i.e., `self`. I assume this is because I've made an egregious logic error, or perhaps more specifically, sharingCancelable is cancelled before it should be.

I'm also not sure if I needed to add those three sharing members to the Rx, but it seems like they needed to be persisted to the Rx.

### DIFF
--- a/monadic-rx/shared/src/main/scala/mhtml/rx.scala
+++ b/monadic-rx/shared/src/main/scala/mhtml/rx.scala
@@ -156,11 +156,11 @@ case class RxImpureOps[+A](self: Rx[A]) extends AnyVal {
   def foreach(effect: A => Unit): Cancelable = Rx.run(self)(effect)
 
   /**
-    * Memoizes this `Rx` using an internal `Var`. This is primarily
-    * useful for optimizing an Rx graph, so that this `Rx` is only
-    * computed once.
+    * Memoizes this `Rx` using an internal `Var`. This is only
+    * useful for optimizing an Rx graph, so that values generated
+    * by this `Rx` are computed only once and shared between all executions.
     */
-  @deprecated("This will eventually be made private and used under-the-hood, automatically.")
+  @deprecated("This will eventually be made private and used under-the-hood, automatically.", "1.0.0")
   def sharing: Rx[A] = Rx.Sharing[A](self)
 }
 
@@ -180,7 +180,7 @@ object Rx {
   final case class Sharing [A]        (self: Rx[A])                                 extends Rx[A] {
     // Should be Var[A], but gives GADT Skolem bug:
     protected[Rx] val sharingMemo: Var[Any] = Var(None)
-    protected[Rx] var isSharing = false
+    protected[Rx] def isSharing = !(sharingCancelable == Cancelable.empty)
     protected[Rx] var sharingCancelable: Cancelable = Cancelable.empty
   }
 
@@ -264,21 +264,18 @@ object Rx {
         Cancelable { () => cc.cancel; self.imitating = false }
       } else run(other)(effect)
 
-    case shareRx @ Sharing(self) =>
-      if (!shareRx.isSharing) {
-        shareRx.sharingCancelable = run(self){ x =>
-          shareRx.sharingMemo := x
-        }
-        shareRx.isSharing = true
+    case rx @ Sharing(self) =>
+      if (!rx.isSharing) {
+        rx.sharingCancelable = run(self)(rx.sharingMemo.:=)
       }
-      val foreachCancelable = shareRx.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
-      Cancelable { () => {
+      val foreachCancelable = rx.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
+      Cancelable { () =>
         foreachCancelable.cancel
-        if (shareRx.sharingMemo.subscribers.isEmpty) {
-          shareRx.sharingCancelable.cancel
-          shareRx.isSharing = false
+        if (rx.sharingMemo.subscribers.isEmpty) {
+          rx.sharingCancelable.cancel
+          rx.sharingCancelable = Cancelable.empty
         }
-      }}
+      }
 
     case leaf: Var[A] =>
       leaf.foreach(effect)
--- a/monadic-rx/shared/src/main/scala/mhtml/rx.scala
+++ b/monadic-rx/shared/src/main/scala/mhtml/rx.scala
@@ -142,7 +142,18 @@ sealed trait Rx[+A] { self =>
    */
   def sampleOn[B](other: Rx[B]): Rx[A] = SampleOn[A, B](this, other)
 
+  /**
+    * Memoizes this `Rx` using an internal `Var`. This is primarily
+    * useful for optimizing an Rx graph, so that this `Rx` is only
+    * computed once.
+    */
+  def sharing: Rx[A] = Sharing[A](this)
+
   val impure: RxImpureOps[A] = RxImpureOps[A](this)
+
+  protected var isSharing = false
+  protected var sharingMemo: Option[Any] = None
+  protected var sharingCancelable: Cancelable = Cancelable.empty
 }
 
 case class RxImpureOps[+A](self: Rx[A]) extends AnyVal {
@@ -169,6 +180,7 @@ object Rx {
   final case class Collect [A, B]     (self: Rx[A], f: PartialFunction[A, B], b: B) extends Rx[B]
   final case class SampleOn[A, B]     (self: Rx[A], other: Rx[B])                   extends Rx[A]
   final case class Imitate [A]        (self: Var[A], other: Rx[A])                  extends Rx[A]
+  final case class Sharing [A]        (self: Rx[A])                                 extends Rx[A]
 
   /**
    * The `impure.foreach` interpreter. Traverses the `Rx` tree and registers
@@ -249,6 +261,18 @@ object Rx {
         }
         Cancelable { () => cc.cancel; self.imitating = false }
       } else run(other)(effect)
+
+    case Sharing(self) =>
+      if (!self.isSharing) {
+        self.sharingCancelable = self.impure.foreach{ x =>
+          self.sharingMemo = Option(x)
+          println(s"hello from sharing: $x") // DEBUG
+          effect(x)
+        }
+        self.isSharing = true
+      }
+      else self.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
+      self.sharingCancelable
 
     case leaf: Var[A] =>
       leaf.foreach(effect)
--- a/monadic-rx/shared/src/main/scala/mhtml/rx.scala
+++ b/monadic-rx/shared/src/main/scala/mhtml/rx.scala
@@ -176,8 +176,8 @@ object Rx {
   final case class Collect [A, B]     (self: Rx[A], f: PartialFunction[A, B], b: B) extends Rx[B]
   final case class SampleOn[A, B]     (self: Rx[A], other: Rx[B])                   extends Rx[A]
   final case class Imitate [A]        (self: Var[A], other: Rx[A])                  extends Rx[A]
-  final case class Sharing [+A]        (self: Rx[A])                                 extends Rx[A] {
-    protected[Rx] var sharingMemo: Option[A] = None
+  final case class Sharing [A]        (self: Rx[A])                                 extends Rx[A] {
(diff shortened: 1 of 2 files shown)


monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -142,7 +142,18 @@ sealed trait Rx[+A] { self =>
...
+    * Memoizes this `Rx` using an internal `Var`. This is primarily
+    * useful for optimizing an Rx graph, so that this `Rx` is only
+    * computed once.
+    */
+  def sharing: Rx[A] = Sharing[A](this)
+
   val impure: RxImpureOps[A] = RxImpureOps[A](this)
+
+  protected var isSharing = false
+  protected var sharingMemo: Option[Any] = None
```
OlivierBlanvillain: `Option[A]`?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +262,18 @@ object Rx {
...
+    case Sharing(self) =>
+      if (!self.isSharing) {
+        self.sharingCancelable = self.impure.foreach{ x =>
+          self.sharingMemo = Option(x)
+          println(s"hello from sharing: $x") // DEBUG
+          effect(x)
+        }
+        self.isSharing = true
+      }
+      else self.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
```
OlivierBlanvillain: You should be able to `assert(self.sharingMemo.nonEmpty)` here

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +262,18 @@ object Rx {
         Cancelable { () => cc.cancel; self.imitating = false }
       } else run(other)(effect)
 
+    case Sharing(self) =>
+      if (!self.isSharing) {
+        self.sharingCancelable = self.impure.foreach{ x =>
```
OlivierBlanvillain: run(self) { x => ... }

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +262,18 @@ object Rx {
...
+          self.sharingMemo = Option(x)
+          println(s"hello from sharing: $x") // DEBUG
+          effect(x)
+        }
+        self.isSharing = true
+      }
+      else self.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
+      self.sharingCancelable
```
OlivierBlanvillain: I think you need something more complicated than that. Here if you register twice and cancel the second one it also cancels the first one...

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -142,7 +142,18 @@ sealed trait Rx[+A] { self =>
...
+  /**
+    * Memoizes this `Rx` using an internal `Var`. This is primarily
+    * useful for optimizing an Rx graph, so that this `Rx` is only
+    * computed once.
+    */
+  def sharing: Rx[A] = Sharing[A](this)
+
   val impure: RxImpureOps[A] = RxImpureOps[A](this)
+
+  protected var isSharing = false
```
OlivierBlanvillain: I think you only need these on the `Sharing` node, not every Rx

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: I seem to get a variance error if I do anything like that, which I guess makes sense being that Rx isn't invariant.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: Not sure I follow , why is that preferable to `Option.foreach`?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: Ah, right - some form of "reference counting" (or registration counting) I guess. `Var.foreach` seems to do this, but I should probably look more closely. I was originally trying to use a `Var[A]` instead of memoizing an `A` directly, but didn't think I needed it - maybe I'll try that.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
OlivierBlanvillain: I remember seeing that as well, do you also get the error if you move the field to the subclass?

monadic-rx/shared/src/main/scala/mhtml/rx.scala
OlivierBlanvillain: `Option.foreach` will silently do nothing if they are no memoized values. IIUC you should know here, by design, than `sharingMemo` is never `None`

bbarker: Ah, right; I guess `Option.get` would serve both purposes, in that case. I suppose I was erring on the side of leaning on language safety features when I probably don't need to. Maybe that's still good practice generally, but probably not in library calls that are frequently used like here.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
bbarker: Oh, that works! Is this because the case classes extending `Rx` are actually invariant in `A`?

bbarker: @OlivierBlanvillain Well, upon compile, I do see the following for `shareRx.sharingMemo = Option(x)`:

```
[error] /home/brandon/workspace/monadic-html/monadic-rx/shared/src/main/scala/mhtml/rx.scala:268: type mismatch;
[error]  found   : x.type (with underlying type A)
[error]  required: ?A10 where type ?A10 <: A (this is a GADT skolem)
[error]           shareRx.sharingMemo = Option(x)
```

I looked into this a bit but am possibly too sleep deprived and/or lacking in knowledge to know if an immediate solution (even seeing your prior comment about a similar error in `Imitate`.

When I tried out:

```scala
    case shareRxGADT @ Sharing(self) =>
      val shareRx: Sharing[A] = shareRxGADT
      if (!shareRx.isSharing) {
```

```
[error]  found   : mhtml.Rx.Sharing[?A10] where type ?A10 <: A (this is a GADT skolem)
[error]  required: mhtml.Rx.Sharing[A]
[error] Note: ?A10 <: A, but class Sharing is invariant in type A.
[error] You may wish to define A as +A instead. (SLS 4.5)
[error]       val shareRx: Sharing[A] = shareRxGADT
[error]                                 ^
[error] one error found
```

So it looks like we could avoid this error, but we'll be back to using `Option[Any]`. I pushed what I have for now.

OlivierBlanvillain: This looks like a pattern matching bug. Given a `Rx[A]` that is an instance of `Sharing[A](self)` it's perfectly valid to put a `Some[A]` inside the sharing. But somehow `scalac` doesn't know than these two `A`s are the same...

OlivierBlanvillain: When I wrote [these lines](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-rx/shared/src/main/scala/mhtml/rx.scala#L218-L219) I remember trying it out in Dotty, which does the correct thing in this case :smile:

bbarker:
@OlivierBlanvillain One thing I want to point out in the test is this line:

```scala
assert(count == 6) //FIXME: should be 4?
```
This seems wrong to me.

Once we get this all fixed up, rather than merging directly, I can squish commits and open a new PR.

bbarker:
Oh, it appears the extra 2 come from the calls to `impure.value`

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -154,6 +154,14 @@ case class RxImpureOps[+A](self: Rx[A]) extends AnyVal {
    * If you use this in your code, you are probably doing in wrong.
    */
   def foreach(effect: A => Unit): Cancelable = Rx.run(self)(effect)
+
+  /**
+    * Memoizes this `Rx` using an internal `Var`. This is primarily
```
OlivierBlanvillain: `s/primarily/*only*/`

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -154,6 +154,14 @@ case class RxImpureOps[+A](self: Rx[A]) extends AnyVal {
...
+    * useful for optimizing an Rx graph, so that this `Rx` is only
```
OlivierBlanvillain: , so that values generated by this `Rx` are computed only once and shared between all executions.

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +264,22 @@ object Rx {
         Cancelable { () => cc.cancel; self.imitating = false }
       } else run(other)(effect)
 
+    case shareRx @ Sharing(self) =>
```
OlivierBlanvillain: It's a small detail but I think that when using `@` like this there is a conversion to shadow the original name such to give it a more precise type:

```scala
rx match {
   ...
   case rx @ Sharing(self) =>
}
```

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +264,22 @@ object Rx {
...
+      if (!shareRx.isSharing) {
+        shareRx.sharingCancelable = run(self){ x =>
+          shareRx.sharingMemo := x
+        }
+        shareRx.isSharing = true
+      }
+      val foreachCancelable = shareRx.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
+      Cancelable { () => {
```
OlivierBlanvillain: Inner braces are optional

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +264,22 @@ object Rx {
         Cancelable { () => cc.cancel; self.imitating = false }
       } else run(other)(effect)
 
+    case shareRx @ Sharing(self) =>
+      if (!shareRx.isSharing) {
+        shareRx.sharingCancelable = run(self){ x =>
```
OlivierBlanvillain: `.run(self)(shareRx.sharingMemo.:=)` ^^

monadic-rx/shared/src/main/scala/mhtml/rx.scala
```diff
@@ -250,6 +264,22 @@ object Rx {
...
+          shareRx.sharingMemo := x
+        }
+        shareRx.isSharing = true
+      }
+      val foreachCancelable = shareRx.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
+      Cancelable { () => {
+        foreachCancelable.cancel
+        if (shareRx.sharingMemo.subscribers.isEmpty) {
+          shareRx.sharingCancelable.cancel
```
OlivierBlanvillain: Don't forget to `shareRx.sharingCancelable = Cancelable.empty` here or it might leak something

BTW I wonder if you could get rid of `isSharing` and use `sharingCancelable eq Cancelable.empty` instead.

bbarker: I have to admit I've been lax about worrying about memory leaks on the JVM - maybe it has made me lazy, but I see your point, thanks! I made `isSharing` a method on `Sharing`. A note, for `eq` I got:

```
value eq is not a member of mhtml.Cancelable
[error] Note that Cancelable extends Any, not AnyRef.
[error] Such types can participate in value classes, but instances
[error] cannot appear in singleton types or in reference comparisons.
[error]     protected[Rx] def isSharing = !(sharingCancelable eq Cancelable.empty)
```

`==` seems to work.

OlivierBlanvillain:
LGTM :+1: Thanks for the PR! (No need to squash, the history makes perfect sense here :smile:)

PR #85, opened 2017-11-07
renaming impure.foreach to impure.run

bbarker:


OlivierBlanvillain:
:+1:

I was actually thinking about dropping the `impure` prefix altogether, WDYT?

bbarker:
Personally, I'd be happy to leave it where it is; I don't think it should feel quite as natural to use as the rest of the API (not in the sense of it being difficult to use, but just that seeing 'impure' before it will make a newcomer think twice).

OlivierBlanvillain:
Alright, thanks for the PR!

PR #86, opened 2017-12-12
[WIP] Optimizer planning

bbarker:
This is a WIP, still need to (and appreciate any help or suggestions for):
- [x] more flatMap examples
- [x] merge example
- [ ] composition examples: tail-shareable mixed with always-shareable
- [ ] quick look at imitate example
- [ ] algorithmic considerations


monadic-rx/shared/src/main/scala/mhtml/Optimizer.md
```diff
@@ -0,0 +1,191 @@
...
+(the diagrams get a bit messy to draw, but I have have them on a white board .
+
+
+## Algorithmic Considerations
+
+**Question**: since we already have asserted that:
+> When combined with a `fold` operator, `flatMap` is problematic: 
+it either leaks memory or breaks referential transparency.
+
+should we detect this and either warn or throw an error?
```
OlivierBlanvillain: This is not the case with mhtml flatMap's due to it's weird "registration" semantic, that is, the current implementation is both memory safe and RT.

What makes this possible is that flatMap is defined such that when it's upstream Rx updates it *cancels* every that did before, call it's `f: A => Rx[B]` with the new `A` and *registers* to the newly returned `Rx[B]`.

The consequence is that if you do something like the following:

```scala
unstream.flatMap(x => other.fold()())
```

Every-time `unstream` updates, a fresh new `fold` is started from on top of the current value of `other`. I hope this makes it obvious why `flatMap` inherits the shareability of whatever its `f` returns :smile:

monadic-rx/shared/src/main/scala/mhtml/Optimizer.md
```diff
@@ -0,0 +1,227 @@
...
+    val numbers: Rx[Int]
+    val folded: Rx[Int] = numbers.foldp(0)(_ + _)
+    val foldedFm: Rx[Int] = numbers.flatMap(x => folded)
+    // numbers      => 1  2 1  1 3 ...
+    // folded       => •1 3 4  5 8 ...
+    // foldedFm1    => •1 3 4  5 8 ...
+    // foldedFm2    =>      •4 5 8 ...  
+    ```
+
+This suggests we should check to see if an `Rx` is defined (**how?**) as this
```
bbarker: @OlivierBlanvillain Thanks for your previous comments, it is a big help. I'm curious now about how we can check this situation, where `folded` is already defined.

Also I'm still most immediately curious about how `merge` isn't always-shareable.

OlivierBlanvillain: Everything is RT, so having `folded` "already defined" or "defined on the fly" cannot make a different (by definition of RT)! I'm think if try it out in a test you will get the following outcome from this example:

```scala
val numbers: Rx[Int]
val folded: Rx[Int] = numbers.foldp(0)(_ + _)
val foldedFm: Rx[Int] = numbers.flatMap(x => folded)
// numbers      =>  1  2  1  1 3 ...
// folded       => •1  3  4  5 8 ...
// foldedFm1    => •1  3  4  5 8 ...
// foldedFm2    =>       •1  2 5 ...  
```

The problem about `merge` comes from the fact that is always uses the latest value from the left hand side Rx as it's first emitted value. For instance, in this example, merged2 is registered after r2 emitted 4 but will start with 8, the latest value from r1.

```scala
val r1: Rx[Int]
val r2: Rx[Int]
val merged: Rx[Int] = r1.merge(r2)
// r1      =>  0    8        3 ...
// r2      =>  1      4    3   ...
// merged1 =>    •0 8 4    3 3 ...
// merged2 =>           •8 3 3 ...
```

You can see that `8 3 3` is not a sub sequence of `0 8 4 3 3`, but `3 3` is, so merge is tail sharable.

bbarker: Ah right, of course - that sort of comes back to the whole point of all this.

For merge, this makes me realize/remember that an Rx must always have a value to emit (I think), so I need to fix some of the other examples a bit to reflect that.

bbarker:
@OlivierBlanvillain When you discuss the example of `val out = rx1.dropIf(_ % 2 == 0)(-1).merge(rx2.dropIf(_ % 2 == 0)(-1))`, are you assuming the `dropIf` rxs may be registered at different times than the `merge` rx? If so, I can easily see how this could give rise to issues, though in the example I pushed where they are all at the same time, I do not see how it can be any worse than tail-shareable. Still, I guess we must always assume "non-temporally transparency", but I am just curious.

OlivierBlanvillain:
I just tried, I could't make an trace for that example that would be worse than tail-shareable. I must have messed up at the time, which means it might compose better than I thought :smile:

> are you assuming the dropIf rxs may be registered at different times than the merge rx?

That would be impossible, `out` is a "Rx expression" that can only be interpreted at a single point of time, as a whole. There is no way to "register" a sub expression of a Rx and use the result as part of a bigger "Rx expression": it would break RT.

bbarker:
Right, I get that different registration points can't exist for this particular expression contained in `out`; maybe it gets more complicated when we break down the expressions individually, which would allow certain time delays to happen between the registration of the various rxs, which is what I was thinking when I said non-temporally transparent (which I guess is just another way of saying not always-shareable).

>  There is no way to "register" a sub expression of a Rx and use the result as part of a bigger "Rx expression": it would break RT.

OK, I understand the sentiment but I think I'm still missing something. Isn't that essentially what we are doing with showing different registration points in these marble diagrams? I suppose one way to do that in code would be to have different threads, with different delays (artificially performed with `Thread.sleep(x)`, but more likely due to user input or network latency in the wild).

OlivierBlanvillain:
That wouldn't be only thing that falls apart if someone uses the current implementation from multiple thread, thankfully that's not an issue on js :smile:

bbarker:
@OlivierBlanvillain Oh right, I proved to myself that the right way to do this in JS is probably using `setTimeout`. I added an example for `merge` in the latest commit - let me know what you think about the slight test restructuring. If that's all good, I can think of two things to do next:

1. Create a test using ScalaCheck (never tried it) showing that tail-shareable composition is still tail-shareable OR prove it OR ignore it for now :)
2. Then move on to algorithmic considerations

monadic-rx/js/src/test/scala/mhtml/RxJsTests.scala
```diff
@@ -0,0 +1,59 @@
...
+    val merged: Rx[Int] = rx1.merge(rx2)
+    var rx1List: List[Int] = Nil
+    var rx2List: List[Int] = Nil
+    var merged1List: List[Int] = Nil
+    var merged2List: List[Int] = Nil
+    val cc1 = rx1.impure.run(n => rx1List = rx1List :+ n)
+    val cc2 = rx2.impure.run(n => rx2List = rx2List :+ n)
+    var ccm1: Cancelable = Cancelable.empty
+    var ccm2: Cancelable = Cancelable.empty
+    setTimeout(1 * delay){
```
OlivierBlanvillain: I'm confused, if you are going to execute these sequentially, what's the point of `setTimeout`? Couldn't you simply move all these statements to the two level and get the same result?

bbarker: Apparently not, I had to modify the results of the asserts to pass as follows (which now appears to indicate *all-shareable():

```scala
      rx1 := 0
      rx2 := 1

      ccm1 = merged.impure.run(n => merged1List = merged1List :+ n)

      rx1 := 8

      rx2 := 4

      ccm2 = merged.impure.run(n => merged2List = merged2List :+ n)

      rx2 := 3

      rx1 := 3

      assert(rx1List == List(0, 0, 8, 3))
      assert(rx2List == List(1, 1, 4, 3))
      assert(merged1List == List(0, 1, 8, 4, 3, 3))
      assert(merged2List == List(8, 4, 3, 3))
```

OlivierBlanvillain: It's really surprising than the two are different...

I just realized than I forgot something about merge, it emits *both* initial values in order, which is why `val out = rx1.dropIf(_ % 2 == 0)(-1).merge(rx2.dropIf(_ % 2 == 0)(-1))` is worse than tail-sharable. Maybe we should change that such that merge only emits the initial value from the left stream, it would at least make things compose more nicely.

The implementation is [here](https://github.com/OlivierBlanvillain/monadic-html/blob/master/monadic-rx/shared/src/main/scala/mhtml/rx.scala#L227-L230). Somehow the effect resisted to the `other` stream should discard the first element. Something like this maybe:

```scala
case Merge(self, other) =>
   val c1 = run(self)(effect)
   var discard = true
   val c2 = run(other) { x =>
      if (discard) discard = false
      else effect(x)
   }
   Cancelable { () => c1.cancel; c2.cancel }
```

bbarker: 1. I agree it is worth considering the sequential issue a bit more - may look at it more tomorrow if not today.

2. >  it emits both initial values in a fixed order (left then right)

This didn't seem to be the case in the added merge example? `merge1List` in the sequential example starts as '0, 8' - if it was as you describe, wouldn't it start as '0, 1'?

3. If the semantics of `merge` were changed as proposed, I suppose it would always be simple enough to initialize any "right-side rx" with a throw-away value, so this should be ok - the user would just need to be aware of the potential gotcha.

bbarker: (of course, it did start as '0, 1' in the non-`setTimeout` version, hmm)

OlivierBlanvillain: > This didn't seem to be the case in the added merge example?

The example is wrong

bbarker: Sorry - could you explain if you mean it is wrong in that it is a bug in the core Rx code, or just that it doesn't meet the current expectations of emitting both left and right "initial" values? Because I believe (unless I typo'd - quite possible) the added example exactly corresponds to the test I added that uses `setTimeout`.

OlivierBlanvillain: Sorry it wasn't clean :smile:. The diagram I draw for merge is wrong according to my understanding of how merge works, I don't understand why the test case you wrote using `setTimeout` behaves this way.

OlivierBlanvillain:
I'm working on a proof (it doesn't look that hard), I just found an interesting example to shows why we need to wait for at least two emitted elements before sharing:

```scala
val numbers: Rx[Int]
val odd: Rx[Int] = numbers.dropIf(_ % 2 == 0)(-1).dropRepeats
// numbers =>    0       0   3   4        5   7 ...
// odd1    =>       •5       3            5   7 ...
// odd2    =>                       •5        7 ...    
```

bbarker:
> (it doesn't look that hard)
I'm glad you think so! I wouldn't mind trying it but I'm sure it would be slow going for me anyway for more than one reason :smile: 

That's certainly an interesting example, particularly since `dropRepeats` is always-shareable.

OlivierBlanvillain:
Here what I have, it's not very polished but I think it works.

Definition: a `Rx` is tail-shareable if any two interpretation *i1* and *i2* of that `Rx` are equivalent after they emitted their first value. Two interpretation are equivalent if there is no observable difference between canceling *i1* or canceling *i2*.


Lemma 1: Values emitted by interpretations of `Rx` composed of `map`, `merge`, `collect` and `Vars` are either 1. the first value of that interpretation or 2. triggered by an update from one of the `Vars`.

Proof: by structural induction on `Rx` with the above structure. `Vars` form the base case and `Map`, `Merge` and `Collect` the inductive case. We inspect the implementation:

- The of interpretation `Vars` forwards to the underlying foreach which implements what is stated in the Lemma.

    case leaf: Var[A] =>
      leaf.foreach(effect)

- `Map` interprets the underlying `Rx` with a pure function: holds by induction hypothesis

    case Map(self, f) =>
      run(self)(x => effect(f(x)))

- `Merge` discards the first value of the other `Rx`, then holds by induction hypothesis

    case Merge(self, other) =>
      val c1 = run(self)(effect)
      var discard = true
      val c2 = run(other) { x =>
         if (discard) discard = false
         else effect(x)
      }
      Cancelable { () => c1.cancel; c2.cancel

- `Collect` first value is either fallback or comes from the underlying `Rx`m then it holds by induction hypothesis

    case Collect(self, f, fallback) =>
      var first = true
      run(self) { a =>
        val out =
          if (f.isDefinedAt(a))
            effect(f(a))
          else if(first)
            effect(fallback)
          else ()
        first = false
        out
      }


Lemma 2: The interpretation of `Rx` composed of `map`, `merge`, `collect` and `Vars` is either pure (no var/mutation) or effectively pure after emitting the first value (it has two states, one to emit the first value and another one for all subsequent values).

Proof: by structural induction on `Rx` with the above structure. `Vars` form the base case and `Map`, `Merge` and `Collect` the inductive case. We inspect the implementation:

- `Vars`: no var/mutation (one also has to look inside the implementation of foreach)

    case leaf: Var[A] =>
      leaf.foreach(effect)

- `Map`: no var/mutation

    case Map(self, f) =>
      run(self)(x => effect(f(x)))

- `Merge`: discard stays false after the first element

    case Merge(self, other) =>
      val c1 = run(self)(effect)
      var discard = true
      val c2 = run(other) { x =>
         if (discard) discard = false
         else effect(x)
      }
      Cancelable { () => c1.cancel; c2.cancel

- `Collect`: first stays false after the first element

    case Collect(self, f, fallback) =>
      var first = true
      run(self) { a =>
        val out =
          if (f.isDefinedAt(a))
            effect(f(a))
          else if(first)
            effect(fallback)
          else ()
        first = false
        out
      }

Theorem: `Rx` composed of `map`, `merge`, collect` and `Vars` are tail-shareable.

Proof: Given two interpretation *i1* and *i2* of such `Rx`, consider any non first value emitted by *i1*. By Lemma 1, this value is triggered by an update from one of the `Vars`, which by constructions must necessarily be at leaf position. By Lemma 2, both interpretation are effectively pure, meaning that *i2* will necessarily produce the same result *i1*. In other words, *i1* and *i2* are equivalent, and the `Rx` is tail-shareable.

bbarker:
@OlivierBlanvillain proof looks good afaikt, I read it on the way down but then got pretty busy/lazy with other things during the holiday :). Just posting a few of my notes here before continuing on:

1. `dropIf`/`keepIf` simply call `collect`, so we get them for free.
2. When you have time and if it is easy, could you add the proof to the OptimizerPlanning.md file in this PR (sorry, I can't copy your unrendered markdown easily).
3. > I think instead if tail-shareability we should use another property to decide when to switch into "shared" mode (a switch that needs to happens at runtime, after variable number of emitted values). ...

    Nice, for this, once the proof is finalized, this means we could simply check to see if the given Rx still maintains/uses internal state, and if not, set it to sharing. One way to do this (not sure if the best way, or even if it works for sure) would be to change `Sharing` to be a trait, then have all Rx's that may be eventually shareable extend it, and use the members of `Sharing` appropriately. However, at the very least that might result in (non-trivial) ode duplication and make things messier, so ideally we could instead do something like:

    ```scala
     trait Stateful {
        // Assume using state initially to be safe
        usingState = true
     }

     // Change Merge to exted Stateful and change Sharing to require Stateful parameter
     final case class Merge   [A, B >: A](self: Rx[A], other: Rx[B])                   extends Rx[B] with Stateful
     final case class Sharing [A]        (self: Rx[A] with Stateful)    

    //Just added an if/else at the outer level to existing Sharing implementation
    case rx @ Sharing(self) =>
      if (self.usingState)  run(self)(effect)
     else {
        if (!rx.isSharing) {
          rx.sharingCancelable = run(self)(rx.sharingMemo.:=)
        }
        val foreachCancelable = rx.sharingMemo.foreach(x => effect(x.asInstanceOf[A]))
        Cancelable { () =>
          foreachCancelable.cancel
          if (rx.sharingMemo.subscribers.isEmpty) {
            rx.sharingCancelable.cancel
            rx.sharingCancelable = Cancelable.empty
          }
        }
      }

    case Merge(self, other) =>
      val c1 = run(self)(effect)
      usingState = true
      val c2 = run(other) { x =>
         if (usingState) usingState = false
         else effect(x)
      }
      Cancelable { () => c1.cancel; c2.cancel }
   ```

bbarker:
@OlivierBlanvillain 

Here is a short sketch of the remaining cases, handled under a modified version of the proof where we only require that the `Rx` is eventually state-free ("effectively pure"); certainly this is far from formal, but if it looks good enough, maybe lets implement and clean up the proof later?

- `zip` - Appears to be eventually shareable, depeding on its underlying `Rx`s: in the implementation of `Zip`, `v1` and `v2` are stateful representations of the values in the underlying `Rx`s `self` and `other`: only after both `self` and `other` themselves have become eventually shareable, does the zipped `Rx` become shareable.
- `dropRepeats` is, perhaps surprisingly, never shareable, since it always depends on its previous values and must track these; however, this is hopefully not much of a problem, but does imply that users should make use of `dropRepeats` as late as possible: `.sharing.dropRepeats` instead of `.dropRepeats.sharing`
- `sampleOn` is similar to `zip`: `currentA` maintains state that reflects the value of the `self` `Rx`, and has no other intracies, therefore `SampleOn` is shareable once both the `self` and `other` `Rx`s are shareable.

Regarding `flatmap`, as discussed it maintains no persistent state, but is not necessarily effetively pure as it depends not only on an underlying `Rx`, but on the emitted `Rx` type. So no guarantees can be made, though it is possible we may be able to optimize sharing if we can dynamically change it each time the shareability of the emitted `Rx` class changes. However, we'll probably want to think about how to implement that later after implementing initial optimization for the simpler cases.

OlivierBlanvillain:
@bbarker I agree on everything you said expect the `dropRepeats` case. I have the intuition than as soon as the previous value hold inside the `dropRepeats` is clean from any "initial propagation" value, then its becomes shareable. If you have a diagram in mind please write it down, it's easier to discuss with a diagram at hand :)

bbarker:
@OlivierBlanvillain I see what you mean now; though state is maintained, it doesn't affect the output value, other than potentially at initialization depending on the underlying Rx. It does affect when the Rx has an effect, but this will strictly depend on the underlying Rx, so it will be at the same time between shared Rx's.

OlivierBlanvillain:
@bbarker Yes, that's exactly it!

bbarker:
@OlivierBlanvillain One issue is that having sharing at every node might generate unwanted overhead for large Rx graphs; ideally, we could bypass the issue if the interpreter could find out when an Rx node in a graph has multiple dependent Rxs; but it looks like there is probably no way to do this, without adding additional overhead to the Rx, so probably not worthwhile (at least in the first pass). Though the overhead might be minimal, something like `subscribers` for `Var`, but only a reference count:

```scala
object Rx {
  //...
  var refCount = 0

  // Then each rx method can increment like:
  def collect[B](f: PartialFunction[A, B])(b: B): Rx[B] = {this.refCount+=1; Collect[A, B](this, f, b)}
}
```

However, I'm not exactly sure how to deal with decrements to `refCount`, which I guess would only happen whenever a particular Rx is gc'd.

OlivierBlanvillain:
I'm not sure I follow, could you give a concrete example?

bbarker:
@OlivierBlanvillain Something like this, where each dashed, black line (other than the first, which is a `Var`) is an ordinary Rx that can be shared (shared = blue circles). If the optimizer simply places them anywhere possible, then we wind up with something like the right figure, I think, where ideally, we only need something like the left.
image

OlivierBlanvillain:
I think the algorithm I have in mind would do the optimal thing in your example. One implementation would be to have a global mutable map from node to interpretation (modulo the sharability logic), whenever an edge is interpreted the global map is first checked for shareable instances, if none is found, then a new interpretation is created for that node and added to the map.

- Tree the user wrote: (`|` edges are rx nodes)

```
 var
  |
  |
/   \
|   |
|   |
```

- 1st interpretation:

```
 var
  |
  |x
/  
|  
|y 
run on the left leaf
```

- 2nd interpretation: (`!` are edges shared with previous interpretations)

```
  !x
    \
    |
    |z
   run on the right left
```

- 3rd interpretation:

```
!y
run on the left leaf
```

- 4th interpretation:

```
    !z
run on the right leaf
```

This means that the runtime actually has a single "interpreted" copy of the graph:

```
 var
  |
  |x
/   \
|   |
|y  |z
```

Where x y and z have more than one callback attached to them.

bbarker:
This sounds good to me in general, but would this be prone to memory leaks? I assume we'll have a mechanism from removing entries from the mutable map when there are zero callbacks, but more on that below:

Drilling down to the details a bit:

> whenever an edge is interpreted the global map is first checked for shareable instances

There still needs to be a way to see if we even need to create a shareable instance, i.e., how does the first or second interpretation in the example above know that `x` (for instance) has more than one callback? What are callbacks in this case (I assume these are not something currently in monadic-rx that go by another name, unless we mean the `effect`s)?

OlivierBlanvillain:
@bbarker This global mutable map thing is just the first idea than came to my mind, I didn't really think it through. Let me try to answer your questions (brainstorming):

> but would this be prone to memory leaks

This should be fine, there is already a mechanism in place to deal with that, I don't see why it couldn't be extended to this sharing thing.

> What are callbacks in this case

Yes, I'm talking about what I've called `effect` in the current implementation.

> There still needs to be a way to see if we even need to create a shareable instance

Indeed this is something that I completely overlooked. The naive solution is to create a dummy shareable node for every actual node, but this might result in quite a bit of wasted resources. (I see that this is what you draw on your right hand side diagram, I get it now!)

An alternative that comes to my mind: keep track of the current callback inserted on every node (in a mutable field). Whenever a second traversal hits on a node to be shared, then create the sharing node on the fly, and "path" the previous callback to go thought that new sharing node instead.

bbarker:
> An alternative that comes to my mind:

@OlivierBlanvillain That sounds pretty solid to me; I think the `refCount` variable, mentioned above would still be used in this case; I don't anticipate any use cases where we would have a number of references on the order where integer overflow would occur ... but I guess one never knows, but at least I think we can ignore that issue for now.

I have more questions, but at this point I think it might be more clear to try and code something up to see how it goes, so will try to do that soon.

bbarker:
@OlivierBlanvillain As a quick sanity check before moving on to other Rxs, just wanted to run this by you (it passes Rx tests):

```
    case rx @ Map(self, f) => {
      var refCount = 0
      if (refCount == 0) {
        run(self){x => refCount += 1; effect(f(x))}
      }
      else {
        val shared = rx.impure.sharing
        run(shared)(effect)
      }
    }
```

PR #88, opened 2017-12-13
Upgrade to the latest versions: sbt 1.0.4, Scala 2.12.4, Scala.js 0.6.21

xerial:
We also may need to split the build for Scala 2.12 and 2.11 in TravisCI because `sbt +test` caused OOM in my local machine.


OlivierBlanvillain:
LGTM, thanks!

PR #91, opened 2017-12-22
adding package.json to allow installation of jsdom in unpriv'd envs

bbarker:
"npm install jsdom" would fail with an error for me, otherwise. This seemed to fix it.


OlivierBlanvillain:
LGTM!

PR #92, opened 2018-01-03
mouse-click style change example/test

bbarker:
While the test works, my real example in the wild still doesn't work, so I continue to bug hunt :)


bbarker:
OK, the latest commit (3rd commit, "added failing asserts for styleClick"), which now causes the test to fail, demonstrates the issue I'm having.

bbarker:
OK, found the issue, will push a 4th commit that fixes it: apparently the DOM inserts a space after `visibility:`!

bbarker:
OK, despite that fix, the actual issue remains when tested in a browser. In firefox I see `style=""` and in chromium I just see `style` with no `=` after it. In any case, the style attribute is empty in both cases, and I can't figure out why. In the latest cleanup commit, I removed a slight asymmetry in the code between style and class that I though might be cause the issue, but the result is the same.

bbarker:
Finally, mystery solved (my CSS is not so good ...) I should have been using `display` and not `visibility`

bbarker:
Just a note: the failed Scala 2.11 build appears to be due to a temporal packaging issue - it built fine earlier. (Also this made me realize I missed the memo that scala 2.11.12 exists now.)

PR #93, opened 2018-01-03
test that demonstrates attribute (style) string modification

bbarker:
This is a failing test currently, and relates to a comment #92 where the style string is modified (spaces added); the end result doesn't affect html, but is a bit surprising and may affect code behavior that doesn't make the assumption that the string changes.


bbarker:
Closing since this is due to the dom impl, not monadic-html.

PR #105, opened 2018-05-03
allowing event handlers that return Rxs

bbarker:
Use case:

```scala
            def actionHandler(ev: MouseEvent): Rx[Unit] = action.command().map {
              case Some(actRes) => actionResult := actRes
              case None => println("no RunResult")
            }
```

If this makes sense, would it be OK to publish an RC2 - I have some production code that may depend on this working, though I guess i could get around it by manually working with `Cancelable`s

### DIFF
--- a/monadic-html/src/main/scala/scala/xml/xml.scala
+++ b/monadic-html/src/main/scala/scala/xml/xml.scala
@@ -1,7 +1,8 @@
 package scala.xml
 
-import language.higherKinds
+import mhtml.Rx
 
+import language.higherKinds
 import scala.annotation.implicitNotFound
 
 // XML Nodes ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


OlivierBlanvillain:
So, you want to use `MouseEvent => Rx[Unit]` an a xml attribute... Could you used `Rx[MouseEvent => Unit] instead?`

monadic-html/src/main/scala/scala/xml/xml.scala
```diff
@@ -146,20 +147,23 @@ The following types are supported:
...
-  @inline implicit def function1AttributeEmbeddable[T]:                     XA[T => Unit]    = null
-  @inline implicit def optionAttributeEmbeddable[C[x] <: Option[x], T: XA]: XA[C[T]]         = null
-  @inline implicit def rxAttributeEmbeddable[C[x] <: mhtml.Rx[x], T: XA]:   XA[C[T]]         = null
+  @inline implicit def noneAttributeEmbeddable:                             XA[None.type]      = null
+  @inline implicit def booleanAttributeEmbeddable:                          XA[Boolean]        = null
+  @inline implicit def stringAttributeEmbeddable:                           XA[String]         = null
+  @inline implicit def textNodeAttributeEmbeddable:                         XA[Text]           = null
+  @inline implicit def function0AttributeEmbeddable:                        XA[() => Unit]     = null
+  @inline implicit def function1AttributeEmbeddable[T]:                     XA[T => Unit]      = null
+  @inline implicit def function0AttributeEmbeddableRx:                      XA[() => Rx[Unit]] = null
```
OlivierBlanvillain: Probably need to be `[C[x] <: mhtml.Rx[x]: XA[() => C[Unit]]` like the other one, or scalac won't consider that implicit for `() => Var[Unit]`.

OlivierBlanvillain:
If `Rx[MouseEvent => Unit]` works for you then I think we don't need to change anything as you can get an instance for that type with `rxAttributeEmbeddable(function1AttributeEmbeddable[MouseEvent])`

bbarker:
Oh, yes, it works, and I'm happy enough with it for now. But this PR would allow more expressions (and shorter expressions) to work that are equivalent, which seems to be a desirable goal.

OlivierBlanvillain:
@bbarker I'm closing your long-standing PR as part of my repo clean-up, I hope you don't mind. Feel free to reopen/ping me if you want to bring anything back to life!

PR #108, opened 2018-10-07
Remove unnecessary use of asInstanceOf in rx.scala

LPTK:


bbarker:
Nice, thanks. We should probably also make an issue to track things that should be changed or need to be changed for dotty.

OlivierBlanvillain:
@LPTK Merci !

@bbarker Besides the uncertain future of xml literals, everything else should hopefully cross compile :smile:

Issue #109, opened 2019-06-11
Scala 2.13 support


antonkulaga:
As Scala 2.13 has been release it will be nice to see it supported by monadic html

kailuowang:
requires the upcoming 0.9.8  release of https://github.com/scala-js/scala-js-dom

gabro:
@kailuowang I've checked today and scala-js-dom 0.9.7 is available for 2.13.0

PR #110, opened 2019-09-21
Add support for Scala 2.13

julienrf:
Support for 2.11 has been disabled because scalacss does not support it.

There is a problem wth XML literals. Has anyone an idea of how to support them?


julienrf:
Fixes #109

examples/target/scala-2.13/examples-opt.js
```diff
@@ -0,0 +1,1214 @@
+(function(){'use strict';
```
esarbe: Can we drop this file?

esarbe:
Can we get this in? Seems alright, no?

examples/target/scala-2.13/examples-opt.js
OlivierBlanvillain: It's used for https://olivierblanvillain.github.io/monadic-html/examples/

saksmt:
Any progress on this?

examples/target/scala-2.13/examples-opt.js
esarbe: Better keep it then.

OlivierBlanvillain:
It's out: https://repo1.maven.org/maven2/in/nvilla/monadic-html_sjs0.6_2.12/0.4.0/

esarbe:
Awesome!

PR #113, opened 2019-12-10
Initial sketch a fanning-out Rx construct

esarbe:
While trying out monadic-html in one of my projects I quickly stumbled over the problem that `Set`s and `List`s will (of course) emit a signal for all subscribers if the list changes. This is an initial sketch that tries to work around that by introducing a `descend` concept. In this initial sketch, it takes a `Rx[Set[T]]` and operates the signals on the level of the elements of the given set.

It pretty much works as expected for `Set`s. I wonder if this can be extended to collections in general and maybe even compound objects.

What do you think?


esarbe:
trying to approach this differently

PR #118, opened 2021-01-25
Disable Sonatype bundle release

olafurpg:


olafurpg:
Looks like the release succeeded this time
```
2021-01-25 09:49:15.340Z  info [SonatypeService] Dropped successfully: innvilla-1035  - (SonatypeService.scala:167)
```

OlivierBlanvillain:
It looks good but I think the tag still got published as a snapshot release and didn't make its way to maven central. Is there something manual to do or is it supposed to be all automated?

olafurpg:
Judging by the version  `0.4.1+0-57d2e133+20210125-0937-SNAPSHOT.pom.asc`, it looks like there was a dirty diff in the git repo when `ci-release` ran. The fix would be to find out what caused that diff and add it to `.gitignore` or commit the changes into the repo. I'm not able to reproduce the diff locally with `+publishLocal` 🤔

olafurpg:
Let's give this a try https://github.com/OlivierBlanvillain/monadic-html/pull/119

PR #121, opened 2021-03-10
prevent eta-expansion in Scala 3

esarbe:


OlivierBlanvillain:
The issue is that if you define cancel with parenthesis it becomes basically unusable from scalac because forgetting parents doesn't trigger a compilation error. Also that is a big breaking change, so if a change is absolutely necessary I would prefer adding an alias instead of changing the existing cancel definition.

What's the situation with dotty? Do you get warnings with the current state?

esarbe:
I get a warning that this will not be supported by dotty anymore. But you are right; it would be quite the breaking change. Let's cross this bridge when we have to.

esarbe:
https://dotty.epfl.ch/docs/reference/changed-features/eta-expansion.html is the current situation with dotty in this regard.

PR #125, opened 2022-11-13
monadic-scalatags

esarbe:
Consider this to be in-progress, a draft.

The absence of types in the xml tags (in combination with my own stupidity) has lead to some bugs in my monadic-html based project in the past. Because of that I've been looking for a way to make my templates a bit more type-safe and during my experimentation I stumbled over scalatags and found that they make a surprisingly good match for monadic-html.

So I've been experimenting with this a while ago and I was wondering if we could bring this into monadic-html somehow.

As mentioned; this is a draft, a sketch, so I'm thankful for any feedback.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,19 +64,20 @@ lazy val `monadic-scalatags` = project
   .dependsOn(`monadic-rx`.js)
   .settings(
     testSettings,
-    libraryDependencies += "com.lihaoyi" %%% "scalatags" % "0.9.3",
-    libraryDependencies += "org.scala-js" %%% "scalajs-dom" % "1.1.0",
+    libraryDependencies += "com.lihaoyi" %%% "scalatags" % "0.12.0",
+    libraryDependencies += "org.scala-js" %%% "scalajs-dom" % scalajsdom,
     libraryDependencies += "org.scalatest" %%% "scalatest" % scalatest % Test
   )
 
 lazy val `examples` = project
   .enablePlugins(ScalaJSPlugin)
-  .dependsOn(`monadic-html`, `monadic-rx-catsJS`)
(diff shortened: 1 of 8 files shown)


OlivierBlanvillain:
Thanks for opening the PR and starting the discussion!

I personally don't like scalatags but I'm not opposed to the idea of having an integration. Could you perhaps translate the examples from the README to see what they would look like? For example, I would like to see how [this one](https://github.com/OlivierBlanvillain/monadic-html/blob/bf6b4743add3afc8b300066024df10772bb5b593/README.md?plain=1#L83-L90) translates to what you are proposing.

I only had a brief look at the diff, but my main concern from the get go is that the APIs would look quite different between the xml and the scalatags version...

esarbe:
Hey, thanks for giving this a think!

That's what the example would look like.

```scala

  val a: Rx[String] = Var("id1")
  val b = Var("foo")
  val c = Var("bar")

  div(
    id := a,
    "Variable 1: ", b, "; variable 2: ", c, "."
  )
```

The API would indeed look quite a bit different. Embedding mhtml tags into scalatag templates looks reasonably simple, but I don't see an easy way of doing it the other way around - that would require changes in the way that `mount` currently works.

I recognize that with scalatags templating will lose one very nice feature - "friendly syntax for frontend developers". I still hope there's room for scalatags templating in mhtml

OlivierBlanvillain:
I was mostly asking about the difference between the `mount(…)` and the `(…).render` syntax, I don’t understand why the scalatags implementation would need to differ here.

Also, looking at the diff again, it seems that you didn’t need to change any of the code that’s already there. Could we possibly envision monadic-scalatags to be a separate library / live in a separate repository?

